Issue #1, opened 2020-03-26
Create a cmake library (in order to have find_package(PoseLib))


pablospe:
Here I have an example:
https://github.com/pablospe/cmake-example-library

This will generate a PoseLibConfig.cmake after "make install" that cmake can find in external projects once the PoseLib library is installed.

PR #8, opened 2020-04-01
C++11 cmake/compilation fix

pmoulon:
Got issues to compile the code with Apple Clang.
- MacOs Catalina 10.15.4 - AppleClang 11.0.3.11030032

The problem at compilation:
```
: error: in-class initialization of non-static data member is a C++11 extension
      [-Werror,-Wc++11-extensions]
    double alpha = 1.0; // either focal length or scale
... -> many more error
```

Proposed solution:
- Add cmake CXX11 feature for the Library and the Benchmark binary.
- CXX11 pre-requirement should propagate in the PoseLib target.

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -7,7 +7,13 @@ add_executable(benchmark
     ${CMAKE_SOURCE_DIR}/re3q3/re3q3/sturm8.cc
 )
 
-target_link_libraries(benchmark ${LIBRARY_NAME} Eigen3::Eigen)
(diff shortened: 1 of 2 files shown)


pablospe:
Thanks for the commit! This is related to issue #1 

One question, besides this fix, does `make install` works well in Apple? (I wasn't able to test). I mean, could you use find_package(...) as indicated in the README?
https://github.com/vlarsson/PoseLib/blob/master/README.md#how-to-use-the-library-as-dependency-in-an-external-project

I haven't closed issue #1, because in Windows `make install` doesn't make sense. If you have any suggestion of what is the correct way to find_package for cmake in Windows...

PR #19, opened 2020-07-09
Hunter package manager (optional use, disabled by default).

pablospe:
Hunter is a CMake driven cross-platform package manager for C/C++. Linux,
Windows, macOS, iOS, Android, Raspberry Pi, etc. It will download the dependencies automatically. More details: https://github.com/cpp-pm/hunter

Hunter is disabled by default. To activate, add `-DHUNTER_ENABLED=ON` to cmake:

    > cmake -DHUNTER_ENABLED=ON ..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.13)
 
+# Hunter (package manager)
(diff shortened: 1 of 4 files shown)


pablospe:
Discussed with Viktor, we don't need Hunter.

PR #20, opened 2020-07-09
Separated `scripts/` folder for building scripts.

pablospe:


vlarsson:
It would be nice to also add a script here for building the pybind

pablospe:
I think I could include an option in the scripts that allows to compile also with python (but it has to be set manually):
PYTHON_PACKAGE=ON

Like this example of the library type:
https://github.com/vlarsson/PoseLib/pull/20/files#diff-bf2a9d7ae77d5e292b9e591d6d7e4866467f0db615e925458720cea0d2b1abf8R23-R25

PR #24, opened 2021-01-28
Python bindings

ignacio-rocco:
Hi, 

I created some python bindings. I had to add the -fPIC compilation flag to PoseLib (I hope this is ok), so that I could link it with the python binding dynamic lib. 

Bindings are incomplete but the remaining functions can easily be added if there is interest in merging this to master. Similarly with the install instructions. 

The `setup.py` file assumes that the user is using Anaconda and that libPoseLib.a has been copied to the conda lib path. Maybe there is a better option?

Best,
Ignacio


pablospe:
This is great!


> The `setup.py` file assumes that the user is using Anaconda and that libPoseLib.a has been copied to the conda lib path. Maybe there is a better option?

Maybe it would make sense to create directly a conda package?
https://www.underworldcode.org/articles/build-conda-packages/

pybind11/setup.py
```diff
@@ -0,0 +1,23 @@
+from setuptools import setup
+
+from pybind11.setup_helpers import Pybind11Extension, build_ext
+from pybind11 import get_cmake_dir
+
+import sys
+import os
+
+__version__ = "0.0.1"
```
pablospe, line 9: This could be taken from the project definition:
https://github.com/vlarsson/PoseLib/blob/master/CMakeLists.txt#L3

Something similar to [version.h.in](https://github.com/vlarsson/PoseLib/blob/master/PoseLib/version.h.in)

It needs to be compiled anyways, maybe we can have a "setup.py.in" and then when you do "cmake .." the `setup.py` with the correct version is generated. Probably this is overcomplicated :P

pybind11/setup.py
```diff
@@ -0,0 +1,23 @@
...
+    extra_objects=[os.path.join(os.getenv('CONDA_PREFIX'),'lib','libPoseLib.a')]),
+]
+
+setup(
+    name="poselib",
+    include_dirs=[os.path.join(os.getenv('CONDA_PREFIX'),'include'),
+                  os.path.join(os.getenv('CONDA_PREFIX'),'include','eigen3')],
+    ext_modules=ext_modules,
+    cmdclass={"build_ext": build_ext},
+)
```
pablospe, line 23: new line

ignacio-rocco:
> This is great!
> 
> > The `setup.py` file assumes that the user is using Anaconda and that libPoseLib.a has been copied to the conda lib path. Maybe there is a better option?
> 
> Maybe it would make sense to create directly a conda package?
> https://www.underworldcode.org/articles/build-conda-packages/

I guess we could make `setup.py` compile PoseLib [using cmake](https://stackoverflow.com/questions/42585210/extending-setuptools-extension-to-use-cmake-in-setup-py) and then build the python extension. Then we could install using [pip and pointing to the git repo](https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/). I've never done any of these, but we could try.

pybind11/setup.py
ignacio-rocco: In that case cmake might also add the right lib and include paths to `setup.py`. I'm sorry but I don't have much experience with cmake, so I wouldn't know how to write `setup.py.in` correctly.

pybind11/pyposelib.cpp
```diff
@@ -63,4 +63,5 @@ PYBIND11_MODULE(poselib, m)
   m.def("gp4ps", &pose_lib::gp4ps_wrapper);
   m.def("gp4ps_kukelova", &pose_lib::gp4ps_kukelova_wrapper);
   m.def("gp4ps_camposeco", &pose_lib::gp4ps_camposeco_wrapper);
+  m.attr("__version__") = "1.0.0";
```
pablospe, line 66: If it is in the C++, it would be possible to use the variable:
`std::string(POSELIB_VERSION)`

For example:
https://github.com/vlarsson/PoseLib/blob/master/PoseLib/poselib.h.in#L23

But `cmake` needs to be run first.

pybind11/setup.py
pablospe: Yes, we can think about it a bit more, but I believe the right way is to call `cmake` to generate the `setup.py` (from `setup.py.in`) in the building directory, and have something like `make pip-package` rule. Similar as Open3D.

pablospe:
> An alternative would be to use pip and install directly from the git repo. For this, we could make `setup.py` compile PoseLib [using cmake](https://stackoverflow.com/questions/42585210/extending-setuptools-extension-to-use-cmake-in-setup-py) and then build the python extension. Then we could install using [pip and pointing to the git repo](https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/). I've never done any of these, but we could try.

Additionally, it could be upload it to pypi.org:
https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives

pybind11/pyposelib.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <PoseLib/types.h>
+#include <PoseLib/p3p.h>
+#include <PoseLib/gp3p.h>
+#include <PoseLib/gp4ps.h>
+
+
+namespace pose_lib {
+    
+
+std::vector<CameraPose> p3p_wrapper(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X){
```
pablospe, line 17: Just wondering if this could be a lambda function, template or (in the wrongs case) a macro.

ignacio-rocco:
Ok, I think we can first have a working approach using cmake to generate `setup.py` with the correct include paths to eigen3 and poselib headers and the correct path to libPoseLib.a. 

This would be a minimal working version that would require the user to run cmake, and then compile an install the python lib using `python setup.py build`, `python setup.py install`. 

So far I was able to get the eigen3 headers path from cmake and substitute it in setup.py.in. I wasn't able to get the poselib headers path nor the path to libPoselib.a.

@pablospe Could you take care of this? You seem more familiar with cmake than me :) Thanks!

pablospe:
Can you check my changes?
    https://github.com/vlarsson/PoseLib/commits/ignacio-rocco/master

I think the cmake part works, but I can't make the python setup working. If you try:

```
mkdir build && cd build
cmake ..
make pip-package (or make install-pip-package)
```

I get this error:
`    ModuleNotFoundError: No module named 'pybind11.setup_helpers'`

ignacio-rocco:
For cmake, I was able to compile both PoseLib and the pip wheel package by doing:
```
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=../install ..
cmake --build . --target install -j 8
make pip-package
```
The error you're getting seems to indicate that you don't have the pybind11 python package installed. Can you open the python interpreter and run: `import pybind11`, followed by `pybind11.__file__`?

pablospe:
I manage to install properly pybind11, and your example works. Now you need to add `-DPYTHON_PACKAGE=ON` to the cmake configuration, making the python compilation optional.

```
cmake -S . -B _build/ -DPYTHON_PACKAGE=ON -DCMAKE_INSTALL_PREFIX=_install
cmake --build _build/ --target install -j 8
cmake --build _build/ --target pip-package
cmake --build _build/ --target install-pip-package
```
(this should also work in Windows, instead of typing `make pip-package`)

pablospe:
I don't see my last chances in the PR. Did you update the branch last time? Otherwise you can see my commits here:
https://github.com/vlarsson/PoseLib/commits/ignacio-rocco/master

pablospe:
Also check: 
https://github.com/vlarsson/PoseLib/pull/24/files#r566737944

ignacio-rocco:
It's working for me as well. I added some printing capabilities to the CameraPose object in python.

I guess now we can improve the binding file `pyposelib.cpp`. 

Pending stuff:

- Possible use lambda functions as wrappers
- Add function arguments [as shown here](https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments)
- Maybe also expose the problem generator to python? and add some tests? (not really needed I guess)

pablospe:
@vlarsson can you also give a look to this PR? Look good to me.

pablospe (approved):

ignacio-rocco:
I took care of the most important points of the list from the previous message:

- Incorporate all the solvers in PoseLib
- Add function arguments

I'm not planning to add anything else. Do you think it's ready for merging? Anything else you want to add?

vlarsson (approved):

vlarsson:
Thanks a lot for setting this up! I think it looks good. I have not had the time to test the bindings but I quickly checked that the cpp parts still compile work for me. I am only a python novice and this would have taken me many hours of googling to figure out.

PR #33, opened 2022-01-11
Robust

vlarsson:
New release with major (breaking) changes.

* Implementation of robust estimators via LO-RANSAC
* Change CameraPose (use quaternion and removed alpha)
* Changed namespace from pose_lib to poselib
* Updated pybinds


.clang-format
```diff
@@ -0,0 +1,11 @@
...
+BasedOnStyle: LLVM
+AlignAfterOpenBracket: Align
+ColumnLimit: '120'
+FixNamespaceComments: 'true'
+IndentWidth: '4'
+Language: Cpp
+PointerAlignment: Right
+TabWidth: '4'
+UseTab: Never
+...
```
pablospe, line 11: What about these options (this is just some suggestions, feel free to ignore):

BasedOnStyle: Google
IncludeBlocks: Preserve
SortIncludes: true
IncludeBlocks: Regroup
IndentWidth: '2'
TabWidth: '2'
ColumnLimit: '100'

pablospe: especially these two:
```
SortIncludes: true
IncludeBlocks: Regroup
```

vlarsson: Sure, I can include the last two. Otherwise I like the setup I have in the other clang-format with 4 space indentation. I will update this and push it.

vlarsson: Fixed in 49880a3

pablospe: sounds good!

PR #40, opened 2022-02-10
Fix a bug of matrix operation in jacobian_impl.h

xuyongzhi:


vlarsson:
Thanks a lot for finding this bug!

xuyongzhi:
My pleasure. Thanks for sharing your wonderful work!

Issue #50, opened 2022-07-26
Build error in robust/sampling - size_t not declared


v-pnk:
Hi and thank you for the code!
I was not able to build the library recently, getting `error: ‘size_t’ was not declared in this scope` for [sampling.h](https://github.com/vlarsson/PoseLib/blob/master/PoseLib/robust/sampling.h) and [sampling.cc](https://github.com/vlarsson/PoseLib/blob/master/PoseLib/robust/sampling.cc).

Probably something in my setup (Ubuntu 22.04, CMake 3.22.1, GCC 11.2.0) or installed libraries is causing the issue, as it seems to work for everyone else. However, adding `#include "PoseLib/types.h"` to [sampling.h](https://github.com/vlarsson/PoseLib/blob/master/PoseLib/robust/sampling.h) fixes the problem for me.

vlarsson:
Interesting. Maybe more modern gcc variants are more strict, I am currently using 9.4.  I never thought about it before since many headers pull in size_t by default. You could add "#include <cstddef>" to sampling.h which I think should resolve it.

pablospe:
I could replicate the issue using docker. It seems we need a fix for this.

pablospe:
Proposed fix from @v-pnk 
https://github.com/vlarsson/PoseLib/pull/51

I will create another PR with a Dockerfile to be able to compile PoseLib inside docker and make these tests easier.

pablospe:
Also check this PR https://github.com/vlarsson/PoseLib/pull/52 (if you want test the docker images).

PR #51, opened 2022-07-29
'size_t' was not declared in this scope

pablospe:


pablospe:
Fixes #50

pablospe:
Tested on ubuntu 20.04 and 22.04 (with docker image, see PR #52).

PR #52, opened 2022-07-29
Add docker support.

pablospe:


pablospe:
Maybe for older version of ubuntu the line in between is needed:

```
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive TZ="Europe/Zurich" apt-get install -y --no-install-recommends tzdata && \
    apt-get install -y libeigen3-dev cmake build-essential
```

I am not going to add this line, just for reference.

pablospe:
And also relaxing the version of Eigen (for ubuntu 20.04):
`find_package(Eigen3 3.3.4 REQUIRED)`

Again, not going to apply this modification, just for reference when trying older version of ubuntu. The added `Dockerfile` is using ubuntu 22.04, and works without modifications.

pablospe:
@vlarsson, if you do `sudo apt install libeigen3-dev` in ubuntu 20.04, the Eigen version is `3.3.4`. Is there any specific reason why to use the `3.4` version? If the alignments aren't a problem in the 3.3.4 version, I would recommend using this version because it will simplify the installation for many people. At least the benchmark seems to work well.

PR #58, opened 2022-12-17
Refactor pybind

vlarsson:
Rewrote the pybind to have a setup.py in the root folder. I find this a bit easier. I added the pybind11, but this could potentially be a submodule. 

@pablospe do you have any opinions on this?


pablospe:
Updating pybind to a newer version will be difficult if it is not a submodule. I would say a submodule here is the best.

CMakeLists.txt
```diff
@@ -20,7 +22,7 @@ if(WITH_BENCHMARK)
 endif()
 
 # python bindings
-option(PYTHON_PACKAGE "Build python package." OFF)
+option(PYTHON_PACKAGE "Build python package." ON)
```
pablospe, line 25: Do you want to make `ON` by default or this was for testing?

pablospe (changes requested):
@vlarsson Thanks to include the pybind as submodule.

I want to check if you want to compile python bindings as default? To simplify compilation of the library, I would leave it as OFF, and `setup.py` should take care of activate this flag. I mean this line:
`option(PYTHON_PACKAGE "Build python package." OFF)`

Maybe also update the following script?`https://github.com/vlarsson/PoseLib/blob/master/scripts/build_python_package.sh`

Btw, you will need to update manually the version in the `setup.py` for each version. I mean the `version="2.0.0"`. I believe this was the reason to have the `setup.py.in` and the `__version__ = "@PROJECT_VERSION@"`. Maybe we can find another workaround, perhaps it could be another file `version.py` that has a `__version__` variable filled out by cmake, and that later on could be imported in `setup.py`, I don't know... (it is like a chicken and egg problem)

pablospe:
@vlarsson Maybe also considering using https://github.com/wjakob/nanobind (same creator of pybind).

vlarsson:
Regarding the version. Indeed this seems to be tricky if we want to allow running "pip install .". Probably easier to write some python code to directly extract the version from the root CMakeLists.txt, compared to having cmake generate some version.py and then import this in python.

pablospe:
It sound good to me, go for the `pip install .`, I was just mentioning you need to remember to modify the version in two places. The only concerns are:
* `option(PYTHON_PACKAGE "Build python package." OFF)`
* and the script: https://github.com/vlarsson/PoseLib/blob/master/scripts/build_python_package.sh

marcusvaltonen:
Regarding the version: the standard way I have seen this being solved is having a VERSION.txt file that both python and CMake reads.

marcusvaltonen:
Furthemore, I would recommend a newer approach, namely using scikit-build-core, which is designed to allow Python packages to be driven from CMake without depending on setuptools. You can then use the `pyproject.toml` approach which is [PEP-518 compatible](https://peps.python.org/pep-0518/).

Some improvements include:
 *   You have to manually copy fixes/additions when they get added to this example (like when Apple Silicon support was added)
 *   Modern editable installs are not supported (scikit-build-core doesn't support them either yet, but probably will soon)
 *   You are depending on setuptools, which can and will change
 *   You are stuck with an all-or-nothing approach to adding cmake/ninja via wheels (scikit-build-core adds these only as needed, so it can be used on BSD, Cygwin, Pyodide, Android, etc)
 *    You are stuck with whatever CMake ships with (scikit-build-core backports FindPython for you)
 
See more [here](https://github.com/pybind/cmake_example).

pablospe:
@marcusvaltonen how do you compare scikit-build-core vs [py-build-cmake](https://github.com/tttapa/py-build-cmake)?

marcusvaltonen:
> @marcusvaltonen how do you compare scikit-build-core vs [py-build-cmake](https://github.com/tttapa/py-build-cmake)?

I haven't tried it myself, but judging from the status of the repo it seems to be in a development phase, e.g., Doxygen and Sphinx support is on the roadmap. Not so many existing packages using it. I am sure it is a nice project, but I would personally chose a more well-established build tool like scikit-build for peace of mind.

vlarsson:
This branch is now redundant as most changes have been merged into dev already (and soon master). See this PR #71 

Specifying the version in a single location is still a TODO. It would be nice, but having to update the version number in CMakeLists.txt and setup.py is not the end of the world.

Regarding moving away from setuptools, I would be hesitant to switch until we encounter any bigger issues with it.

PR #63, opened 2023-03-16
expose the private headers to be public.

B1ueber2y:


PoseLib/CMakeLists.txt
```diff
@@ -80,22 +85,16 @@ set(HEADERS_PUBLIC
...
-set(HEADERS_PRIVATE
-    misc/qep.h
-    misc/univariate.h
-    misc/sturm.h
-    misc/essential.h
-    misc/re3q3.h
-    robust/lm_impl.h
-    robust/jacobian_impl.h
-)
+set(HEADERS_PRIVATE )
```
pablospe, line 97: `robust/lm_impl.h` and `robust/jacobian_impl.h` need to be public too? I wouldn't expose this.

B1ueber2y: In my case I do not need those headers either. I just need to reuse the headers from the misc folder in my own project. But I guess it does not make a big difference to expose those two, or does it?

B1ueber2y: I am fine if we revert the two to be private again

PR #67, opened 2023-06-06
Add a p3p solver from "Revisiting the P3P Problem, CVPR 2023"

S-o-T:
This PR introduce an implementation of p3p solver described in "Y. Ding, J. Yang, V. Larsson, C. Olsson, K. Åström, Revisiting the P3P Problem, CVPR 2023". If you find no issues with correctness/performance, then i can rework this PR to substitute current implementation of p3p (Lambda Twist).

                    Solver    Solutions        Valid     GT found      Runtime
                       p3p      1.72803          100          100    270.85 ns
                  p3p_ding      1.72803          100          100   215.699 ns

### DIFF
--- a/PoseLib/solvers/p3p_common.h
+++ b/PoseLib/solvers/p3p_common.h
@@ -1,0 +1,35 @@
+#ifndef POSELIB_P3P_COMMON_H
+#define POSELIB_P3P_COMMON_H
+
+#include <cmath>
+
+namespace poselib {
+
+// Performs a few newton steps on the equations
+inline void refine_lambda(double &lambda1, double &lambda2, double &lambda3, const double a12, const double a13,
+                          const double a23, const double b12, const double b13, const double b23) {
+
+    for (int iter = 0; iter < 5; ++iter) {
+        double r1 = (lambda1 * lambda1 - 2.0 * lambda1 * lambda2 * b12 + lambda2 * lambda2 - a12);
+        double r2 = (lambda1 * lambda1 - 2.0 * lambda1 * lambda3 * b13 + lambda3 * lambda3 - a13);
+        double r3 = (lambda2 * lambda2 - 2.0 * lambda2 * lambda3 * b23 + lambda3 * lambda3 - a23);
+        if (std::abs(r1) + std::abs(r2) + std::abs(r3) < 1e-10)
+            return;
+        double x11 = lambda1 - lambda2 * b12;
+        double x12 = lambda2 - lambda1 * b12;
+        double x21 = lambda1 - lambda3 * b13;
+        double x23 = lambda3 - lambda1 * b13;
+        double x32 = lambda2 - lambda3 * b23;
+        double x33 = lambda3 - lambda2 * b23;
+        double detJ = 0.5 / (x11 * x23 * x32 + x12 * x21 * x33); // half minus inverse determinant
+        // This uses the closed form of the inverse for the jacobean.
+        // Due to the zero elements this actually becomes quite nice.
+        lambda1 += (-x23 * x32 * r1 - x12 * x33 * r2 + x12 * x23 * r3) * detJ;
+        lambda2 += (-x21 * x33 * r1 + x11 * x33 * r2 - x11 * x23 * r3) * detJ;
+        lambda3 += (x21 * x32 * r1 - x11 * x32 * r2 - x12 * x21 * r3) * detJ;
+    }
+}
+
+} // namespace poselib
+
+#endif // POSELIB_P3P_COMMON_H
--- a/PoseLib/solvers/p3p_ding.cc
+++ b/PoseLib/solvers/p3p_ding.cc
@@ -143,7 +143,7 @@ void compute_pose(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen
 
 int p3p_ding(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X,
              std::vector<CameraPose> *output) {
-    if (output != nullptr) {
+    if (output == nullptr) {
         return 0;
     }
     output->clear();
--- a/PoseLib/solvers/p3p_ding.cc
+++ b/PoseLib/solvers/p3p_ding.cc
@@ -42,7 +42,7 @@ double cubic_trigonometric_solution(const double alpha, const double beta, const
     return 2.0 * I * K - k2 / 3.0;
 }
 
-double cubic_Cardano_solution(const double beta, const double G, const double k2) {
+double cubic_cardano_solution(const double beta, const double G, const double k2) {
     const double M = std::cbrt(-0.5 * beta + sqrt(G));
     const double N = -std::cbrt(0.5 * beta + sqrt(G));
     return M + N - k2 / 3.0;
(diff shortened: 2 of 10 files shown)


tsattler:
Is replacing the solver really worth it? I don't think it will make any run-time difference in RANSAC (as the slowest parts are inlier counting and local optimization). Are there any advantages in terms of robustness?

PoseLib/solvers/p3p_common.h
```diff
@@ -0,0 +1,35 @@
+#ifndef POSELIB_P3P_COMMON_H
+#define POSELIB_P3P_COMMON_H
+
+#include <cmath>
+
+namespace poselib {
+
+// Performs a few newton steps on the equations
+inline void refine_lambda(double &lambda1, double &lambda2, double &lambda3, const double a12, const double a13,
```
pablospe, line 9: Shouldn't this be in a p3p_lamda.h file? I believe that if you are planning to move out, I would create a p3p_lambda.h|cc

PoseLib/solvers/p3p_ding.cc
```diff
@@ -0,0 +1,203 @@
...
+
+double cubic_trigonometric_solution(const double alpha, const double beta, const double k2) {
+    const double H = std::sqrt(-alpha * alpha * alpha / 27.0);
+    const double I = std::sqrt(-alpha / 3.0);
+    const double J = std::acos(-beta / (2.0 * H));
+    const double K = std::cos(J / 3.0);
+    return 2.0 * I * K - k2 / 3.0;
+}
+
+double cubic_Cardano_solution(const double beta, const double G, const double k2) {
```
pablospe, line 45: cubic_cardano_solution (all lower case?)

PoseLib/solvers/p3p_ding.cc
```diff
@@ -0,0 +1,203 @@
...
+    pq[0] = D.col(0);
+    pq[1] = D.row(0);
+
+    return pq;
+}
+
+std::pair<int, std::array<double, 2>> compute_line_conic_intersection(Eigen::Vector3d &l, const double b,
+                                                                      const double m13, const double m23) {
+    std::pair<int, std::array<double, 2>> result;
+    const double cxa = -b * l(1) * l(1) + l(2) * l(2);
```
pablospe, line 106: nit: this is being used in other places, I think you could create a 
`const double l22 = l(2) * l(2);`
and so on...

PoseLib/solvers/p3p_ding.cc
```diff
@@ -0,0 +1,203 @@
...
+    B.col(2) = B.col(0).cross(B.col(1));
+
+    Eigen::Matrix3d R = B * A.inverse();
+    Eigen::Vector3d t = d1 * x[0] - R * X[0];
+    output->emplace_back(R, t);
+}
+
+int p3p_ding(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X,
+             std::vector<CameraPose> *output) {
+    output->clear();
```
pablospe, line 146: check if it isn't nullptr ?

PoseLib/solvers/relpose_8pt.cc
```diff
@@ -31,6 +31,7 @@
 #include "PoseLib/misc/essential.h"
 
 #include <array>
+#include <cassert>
```
pablospe, line 34: needed?

benchmark/benchmark.cc
```diff
@@ -31,30 +31,25 @@ template <typename Solver> BenchmarkResult benchmark(int n_problems, const Probl
         double pose_error = std::numeric_limits<double>::max();
 
         result.solutions_ += sols;
-        // std::cout << "\nGt: " << instance.pose_gt.R << "\n"<< instance.pose_gt.t << "\n";
+        // std::cout << "\nGt: " << instance.pose_gt.R() << "\n"<< instance.pose_gt.t << "\n";
```
pablospe, line 34: Probably we should remove these debug comments? @vlarsson ?

PoseLib/solvers/relpose_8pt.cc
S-o-T: Yes, otherwise gcc fails with following message:

PoseLib/PoseLib/solvers/relpose_8pt.cc:45:5: error: use of undeclared identifier 'assert'
    assert(x1.size() == x2.size());

PoseLib/solvers/p3p_ding.cc
S-o-T: I don't think that it is needed. In my opinion such additional variables does not add into readability nor do they affect performance.

PoseLib/solvers/p3p_common.h
S-o-T: Moving it into separate compilation unit will forbid compile-time optimizations (gcc do perform better when refine_lambda() implementation do available during p3p_ding.cc compilation) and i think that PoseLib does not configure linker to perform lto.

PoseLib/solvers/p3p_ding.cc
```diff
@@ -143,6 +143,9 @@ void compute_pose(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen
 
 int p3p_ding(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X,
              std::vector<CameraPose> *output) {
+    if (!output) {
```
pablospe, line 146: `if (output!= nullptr)`
 maybe more explicit?

tsattler:
> I don't see any issues with that, assuming that proposed solution is more robust and more performant.

Looking at Tab. 3 in the paper, the difference to the Lambda Twist P3P seems marginal: Lambda Twist does not provide a good solution in 11 out of 10M cases, fails in 2 out of 10M cases, and is about 40 ns slower. 
I don't think the new solver will be any more stable than Lambda Twist in practice .

vlarsson:
Thanks @S-o-T for setting this up.

@tsattler while I agree that the improvement is marginal in practice, I don't see any reason for not including it.

pablospe:
I think @tsattler's point is about being the default or not. I also agree that this solver should be included, but I don't know if this should be the default p3p solver, maybe not in this PR (in a future one when we see this is really stable solver) and only included as an extra solver for now to play around?

vlarsson:
Ah yes, this I agree with. We can keep the old solver as the default for now.

benchmark/benchmark.cc
vlarsson: I think it's fine to keep these. They are convenient when you are debugging new solvers :)

vlarsson:
Sorry for the delay. I just want to do some quick comparisons with the code Yaqing wrote for our experiments.

vlarsson:
I did some benchmarking and the solver seems to be very very slightly less stable than the implementation I got from Yaqing. I have no idea why though. I might dig into it a bit more. However, the code is much cleaner, so I will merge this for now, and try to figure it out later.

S-o-T:
@vlarsson Thanks for a review.

Sorry, that i am bringing it up a bit late, but i now see in commit history mentions of enabling usage of Eigen 3.3, so i assume that Poselib currently do not require Eigen 3.4. In such case, the usage of arrays of indices (which is a feature introduced with Eigen 3.4) [during](https://github.com/vlarsson/PoseLib/blob/aa59576b4e6c3577a28c29a956bbaba5ed53ebaa/PoseLib/solvers/p3p_ding.cc#L66) calculation of matrix minors should be reimplemented. If you think that it should be done, i will fix that and open a PR.

vlarsson:
Ah yes, we have reverted to 3.3 in the dev branch (as many people had trouble getting 3.4). If you can open a PR to make it 3.3 compatible that would be great. We should probably merge dev to main soon as well.

PR #75, opened 2023-09-30
Add a basic CI pipeline (.github/workflows/main.yml)

pablospe:
This adds a basic continuous integration. It will compile the code and run the benchmark. Perhaps we want to separate running the benchmark as an optional step.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,42 @@
+name: Compile PoseLib and Run Benchmark
+
+on:
+  pull_request:
+    branches:
+      - '**'
+
+env:
+  BUILD_TYPE: Release
+  BUILD_SHARED_LIBS: OFF # static
+  BUILD_DIR: _build
+  INSTALL_DIR: _install
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


vlarsson:
Looks good to me, but I also don't know anything about GitHub Actions. Perhaps @Parskatt has a better overview.

Parskatt:
Came down with a cold, so currently out of order 🙃 The previous PR looked good.

________________________________

vlarsson:
As far as I can till this only compiles the C++ library (and potentially benchmark). We should make sure PYTHON_PACKAGE is set so that we check that the pybind compiles as well.

pablospe:
I think it would make sense to run each time the benchmark, instead of being optional. It doesn't take so much time actually. I will change, let me know if you disagree.

vlarsson:
There are unit tests in the WIP progress branch https://github.com/PoseLib/PoseLib/tree/camera_models
One option would be to directly integrate the old benchmark into the unit tests (and perhaps retire benchmark/benchmark entirely)

pablospe:
@vlarsson  This PR is ready. Please made another pass, and approve if you agree (remember that now I can't merge it to master without approval).

Click here to see the steps in the pipeline:
image

vlarsson (approved):
LGTM

PR #77, opened 2023-10-03
Check formatting using clang-format

pablospe:


pablospe:
If the code is not formatted with clang, it will fail. So you need to make another push with the code formatted to pass this check

vlarsson (approved):
LGTM

PR #78, opened 2023-10-12
Additional upright solvers

vlarsson:
Additional solvers for absolute pose with known gravity direction. Contains additional overloads for non-aligned gravity directions.


pablospe (approved):
Remember to format the code :)

PR #79, opened 2023-10-17
Disable alignment.h when compiling for modern standards

S-o-T:
Due to possibility to face an error of specialization after instantiation (e.g. https://github.com/colmap/colmap/issues/820) and due cpp >= 17 making such specialization unnecessary (it was removed from colmap https://github.com/colmap/colmap/pull/1947), i suggest to enable alignment.h only when it's needed.

### DIFF
--- a/PoseLib/alignment.h
+++ b/PoseLib/alignment.h
@@ -36,6 +36,14 @@
 #ifndef COLMAP_SRC_UTIL_ALIGNMENT_H_
 #define COLMAP_SRC_UTIL_ALIGNMENT_H_
 
+#include <Eigen/src/Core/util/Macros.h>
+#ifdef _MSVC_LANG
+#define CPP_VERSION _MSVC_LANG
+#else
+#define CPP_VERSION __cplusplus
+#endif
+#if !EIGEN_VERSION_AT_LEAST(3, 4, 0) || CPP_VERSION < 201703L
+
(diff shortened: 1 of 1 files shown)


pablospe:
Why not to remove it completely as in colmap PR?

S-o-T:
Actually, if i understood correctly, the issue is not only standard version, but also Eigen version: we can avoid manual specialization only starting with C++17 **and** Eigen 3.4, so i should add a check for Eigen version also (https://libeigen.gitlab.io/docs/group__TopicStlContainers.html https://web.archive.org/web/20230813133713/https://eigen.tuxfamily.org/index.php?title=3.4). It should not be removed if consumers using older standard and Eigen are to be respected (i am ok with removal). Please, correct me if i am wrong.

As for why it was completely removed in colmap (given that its still [targets C++14](https://github.com/colmap/colmap/blob/ff8842e7d9e985bd0dd87169f61d5aaeb309ab32/CMakeLists.txt#L38) and do [not requires Eigen>=3.4](https://github.com/colmap/colmap/blob/ff8842e7d9e985bd0dd87169f61d5aaeb309ab32/cmake/FindDependencies.cmake#L13)) - i am not sure that i have an answer, maybe Johannes can help us with that.

ahojnnes:
LGTM

PR #80, opened 2023-10-17
Disable -Warray-bounds for bundle.cc due to false positive on GCC>=12

S-o-T:
Due to gcc>=12 raising false positive warning (turned into an error by -Werror) about accessing outside of array bounds (this issue is reported in [Eigen](https://gitlab.com/libeigen/eigen/-/issues/2506) and in [gcc](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106247)), i suggest to suppress such diagnostic in file where it happening and for compilers that are currently affected.


pablospe:
We would need to track this as a github issue, until gcc fixes the problem.

S-o-T:
#81

vlarsson (approved):

PR #86, opened 2023-12-04
Rename uninstall target to be more specific

vlarsson:
Renames the uninstall target so that it's easier to integrate into other projects.

See here https://github.com/colmap/colmap/pull/2288#discussion_r1413180196

### DIFF
--- a/cmake/SetEnv.cmake
+++ b/cmake/SetEnv.cmake
@@ -105,7 +105,7 @@ configure_package_config_file(
 configure_file("${PROJECT_SOURCE_DIR}/cmake/Uninstall.cmake.in"
   "${GENERATED_DIR}/Uninstall.cmake"
   IMMEDIATE @ONLY)
-add_custom_target(uninstall
+add_custom_target(uninstall_poselib
   COMMAND ${CMAKE_COMMAND} -P ${GENERATED_DIR}/Uninstall.cmake)
 
 # Always full RPATH (for shared libraries)


pablospe:
What about checking if the `uninstall` is already defined:
```
# Check if the uninstall target already exists
if(NOT TARGET uninstall)
    # Define the uninstall target
    add_custom_target(uninstall
        COMMAND ...
    )
endif()
```


And add another alternative to include poselib as external project in the `README`. Relevant for this case, which it is:
```
include(FetchContent)
FetchContent_Declare(
    poselib
    GIT_REPOSITORY <URL-of-PoseLib>
    GIT_TAG <desired-commit-or-tag>
    # Other options as necessary
)
FetchContent_MakeAvailable(poselib)
```

This is the default for google test (what [they recommend](https://google.github.io/googletest/quickstart-cmake.html#set-up-a-project) for external project after cmake 3.11). This will download the code during compilation time, avoiding the need to install it, which makes the `uninstall` target irrelevant.

vlarsson:
Sure, naming it uninstall_poselib conditionally is fine with me.

FetchContent sounds convenient and I guess we should promote this in the readme as well then. But from what I understand this also requires the uninstall rename? (if there already is an unsinstall target I mean?)

pablospe:
ok, I will test the `FetchContent` in a separated PR. It is fine with me to simply rename `uninstall` to `uninstall_poselib`.

PR #87, opened 2023-12-04
Bugfix in p4pf pybind

vlarsson:
Previous update to p4pf was not reflected in the pybind.

### DIFF
--- a/PoseLib/solvers/p4pf.cc
+++ b/PoseLib/solvers/p4pf.cc
@@ -46,7 +46,7 @@ int p4pf(const std::vector<Eigen::Vector2d> &x, const std::vector<Eigen::Vector3
 
         for (int i = 0; i < n; ++i) {
             double a = fx[i] / fy[i];
-            double err = std::min(std::abs(a - 1.0), std::abs(1 / a - 1.0));
+            double err = std::max(std::abs(a - 1.0), std::abs(1 / a - 1.0));
             if (err < best_err) {
                 best_err = err;
                 best_ind = i;
--- a/PoseLib/solvers/p4pf.cc
+++ b/PoseLib/solvers/p4pf.cc
@@ -32,12 +32,47 @@
 
 namespace poselib {
 
-int p4pf(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X, std::vector<CameraPose> *output,
+int p4pf(const std::vector<Eigen::Vector2d> &x, const std::vector<Eigen::Vector3d> &X, std::vector<CameraPose> *output,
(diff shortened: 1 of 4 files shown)


pablospe:
yes, good idea. How would you set the `PYTHON_PACKAGE` properly, do you have an example? Let's do it in this PR so we can already test it. The `main.yml` file have to be modified only. Please check here:
https://github.com/PoseLib/PoseLib/blob/master/.github/workflows/main.yml#L34
```
...
      run: |
        cmake \
          -B ${{github.workspace}}/${{env.BUILD_DIR}} \
          -DWITH_BENCHMARK=ON \
          -DCMAKE_BUILD_TYPE=${{env.BUILD_CONFIGURATION}} \
          -DBUILD_SHARED_LIBS=${{env.LIBRARY_TYPE}} \
          -DCMAKE_INSTALL_PREFIX=${{github.workspace}}/${{env.INSTALL_DIR}}
...
```

PoseLib/solvers/p4pf.cc
```diff
@@ -144,9 +179,13 @@ int p4pf(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3
     int n_sols = re3q3::re3q3(coeffs, &solutions);
 
     CameraPose best_pose;
-    double best_focal = 1.0;
-    double best_res = 1.0; // we are probably not interested in solutions above this res anyway
     output->clear();
+    output->reserve(n_sols);
```
pablospe, line 183: wondering if `output->resize` is better here.

pablospe:
> How would you set the `PYTHON_PACKAGE` properly, do you have an example?

Just using `-DPYTHON_PACKAGE=ON`, right?

pablospe:
Now I am thinking better, after we have this PR https://github.com/PoseLib/PoseLib/pull/73 in place, this won't be needed. It would be like a temporal change.

pablospe:
@vlarsson. I made some changes, now it compiles also `pyposelib.cc`.

PoseLib/solvers/p4pf.cc
vlarsson: I was thinking that we could get negative focal lengths here (and thus would not always have n_sol solutions), but I guess it's not the case. I will think more about it and update it.

Parskatt (approved):
LGTM

vlarsson:
I think I could not approve it as I was co-authoring the PR.

pablospe:
ok, let's check in a new PR to see how it goes, and adjust accordantly. I wanted to give more options for approval, rather than being more restrictive.

PR #93, opened 2024-02-14
Refined P3P solver, and a new implementation for rel_pose_upright_3pt

yaqding:
Refined P3P_ding solver. 
Compare with default solver with Newton refinement:
Screenshot from 2024-02-14 22-25-14
without Newton refinement:
Screenshot from 2024-02-14 22-25-43
Performance on real dataset. Slight improvements over the existing solvers. First sequence AUC5 89.57 vs 89.48 and more efficient.
Latest implementation:
Screenshot from 2024-02-14 17-25-23
Default P3P solver:
Screenshot from 2024-02-14 17-25-53
Original p3p_ding by Mark Shachkov:
Screenshot from 2024-02-14 18-07-12

A new implementation for rel_pose_upright_3pt solver based on intersections of conics.
Default solver:
thumbnail_image
New implementation:
Screenshot from 2024-02-14 22-01-52


pablospe:
Check C++ formatting:
https://github.com/PoseLib/PoseLib/actions/runs/7876199471/job/21496196436?pr=92

You can run this line at the top-level in the repo:
https://github.com/PoseLib/PoseLib/actions/runs/7876199471/workflow?pr=92#L31

yaqding:
Thanks! It should be fine now.

S-o-T:
@yaqding thanks for refinement!

I took a quick look over changes and found out that ensuring BC is largest is enough to make current implementation as stable as lambda twist (according to poselib internal benchmark).

Regarding posebench results - i failed to reproduce numbers you posted, all 4 implementation yielded the same AUC5=89.48 on eth3d_130_dusmanu, but it looks a strange given that current implementation is a bit less stable then lambda twist. I will try to find some time to recheck that.

PoseLib/solvers/p3p_ding.cc
```diff
@@ -26,178 +26,207 @@
...
-    const double b = a13 / a23;
+    Eigen::Vector3d X01 = X_copy[0] - X_copy[1];
+    Eigen::Vector3d X02 = X_copy[0] - X_copy[2];
+    Eigen::Vector3d X12 = X_copy[1] - X_copy[2];
+
+    const double a12_copy = X01.squaredNorm();
+    const double a13_copy = X02.squaredNorm();
+    const double a23_copy = X12.squaredNorm();
+
+    std::vector<Eigen::Vector3d> X = X_copy;
```
S-o-T, line 54: Allocating here is not quite optimal, instead, std::array could be used to gain some performance.

yaqding: Thanks! It improves a bit.

yaqding:
@S-o-T The little difference in the posebench is possibly due to the different random samples in different machines. After using std::array instead of std::vector, now the runtime is ~209ns.

PoseLib/solvers/p3p_common.h
```diff
@@ -5,6 +5,86 @@
 
 namespace poselib {
 
+inline double cubic_trigonometric_solution(const double alpha, const double beta, const double k2,
```
vlarsson, line 8: I guess these functions are similar to what exists in https://github.com/PoseLib/PoseLib/blob/master/PoseLib/misc/univariate.cc#L74
?
We should probably figure out what is best and put it here

PoseLib/solvers/p3p_ding.cc
```diff
@@ -26,178 +26,207 @@
...
+    std::array<Eigen::Vector3d,3> x = {x_copy[0], x_copy[1], x_copy[2]};
+
+    double a12;
+    double a13;
+    double a23;
+
+    // Switch X,x so that BC is the largest distance among {X01, X02, X12}
+    if (a12_copy > a13_copy) {
+        if (a12_copy > a23_copy) {
+            x[0] = x_copy[2];
```
pablospe, line 64:
```
            x[0] = x_copy[2];
            x[2] = x_copy[0];
            X[0] = X_copy[2];
            X[2] = X_copy[0];
```

Remove duplicates

PoseLib/solvers/p3p_ding.cc
```diff
@@ -26,178 +26,207 @@
...
-    const double b = a13 / a23;
+    Eigen::Vector3d X01 = X_copy[0] - X_copy[1];
+    Eigen::Vector3d X02 = X_copy[0] - X_copy[2];
+    Eigen::Vector3d X12 = X_copy[1] - X_copy[2];
+
+    const double a12_copy = X01.squaredNorm();
+    const double a13_copy = X02.squaredNorm();
+    const double a23_copy = X12.squaredNorm();
+
+    std::array<Eigen::Vector3d,3> X = {X_copy[0], X_copy[1], X_copy[2]};
```
pablospe, line 54: why is the copy needed? Isn't it a way to swap the columns (efficiently) internally in Eigen?

PoseLib/solvers/p3p_ding.cc
```diff
@@ -26,178 +26,207 @@
...
+            X02 = -X02;
+        } else {
+            a12 = a12_copy;
+            a13 = a13_copy;
+            a23 = a23_copy;
+        }
+    } else if (a13_copy > a23_copy) {
+        x[0] = x_copy[1];
+        x[1] = x_copy[0];
+        X[0] = X_copy[1];
```
pablospe, line 81: also here: `x[0] = x_copy[1];` is twice

PoseLib/solvers/p3p_ding.cc
```diff
@@ -26,178 +26,207 @@
...
+        a12 = a12_copy;
+        a13 = a13_copy;
+        a23 = a23_copy;
+    }
+
+    const double a23d = 1.0 / a23;
+    const double a = a12 * a23d;
+    const double b = a13 * a23d;
 
     const double m12 = x[0].dot(x[1]);
```
pablospe, line 98: wouldn't it be convenient to keep 0-index notation? like have:
```
const double m01 = x[0].dot(x[1]);
```

PoseLib/solvers/relpose_upright_3pt.cc
```diff
@@ -26,85 +26,185 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+// Author: Yaqing Ding (yaq.ding@gmail.com).
```
pablospe, line 29: are you sure you want to keep your email public? :) anyways it is already in this branch. You might need to do a squash and force push, I don't know.

pablospe:
It seems it is still complaining about the formatting.
https://github.com/PoseLib/PoseLib/actions/runs/7927921250/job/21649068111?pr=93

PoseLib/solvers/p3p_ding.cc
yaqding: The last two rows are capital.

PoseLib/solvers/p3p_ding.cc
yaqding: All the parameters renamed .

PoseLib/solvers/p3p_ding.cc
yaqding: I agree, it's not elegant to use additional X_copy, x_copy. The input data is const std::vector<Eigen::Vector3d> &x, it seems that there are no good ways to swap the elements. Or any suggestions?

PoseLib/solvers/p3p_ding.cc
pablospe: ah, thanks!

PoseLib/solvers/relpose_upright_3pt.cc
```diff
@@ -201,7 +188,7 @@ int relpose_upright_3pt(const std::vector<Eigen::Vector3d> &x1, const std::vecto
             }
         }
 
-        if ((n_sols > 0 && G > 0))
+        if ((n_sols > 0 && G))
```
pablospe, line 191: double `((`, needed?

vlarsson (approved):
Looks good to me. We could potentially switch this to be the default p3p solver.

yaqding:
Did minor changes to set as default solver.

vlarsson:
Maybe make the older solver available as p3p_lambdatwist as well.

yaqding:
added now

PoseLib/solvers/relpose_upright_3pt.cc
```diff
@@ -26,85 +26,172 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+// Author: Yaqing Ding.
+// Some of the scripts are based on Mark Shachkov (mark.shachkov@gmail.com) and the Lambda-twist P3P implementation
```
pablospe, line 30: Before merging, I would double-check with Mark two things:
1. Using the code you are mentioning (since this library has BSD library).
2. Adding his email in a public code. I would remove his email.

Note: if it is a total re-implementation I would also rewrite this, perhaps pointing to a publication.

pablospe (changes requested):

PoseLib/solvers/relpose_upright_3pt.cc
S-o-T: @pablospe Thanks for raising that.

Regarding code authorship credits, i believe that point of convergence could be:
```
Authors: Yaqing Ding
         Mark Shachkov
```

As for my email address: i originally put it there, but I think it's a good idea to remove it.

pablospe (approved):

Issue #89, opened 2024-01-10
Adding 6pt relative pose with unknown equal focals + other intrinsic refinement methods


kocurvik:
Hello,

I have a question about potential contributions. For our [recent paper](https://arxiv.org/abs/2311.16304) I have implemented a solver for relative pose with unknown focal length equal for both cameras + its LO into PoseLib. In the paper we propose to perform degeneracy checks for imaginary focal lengths (denoted RFC) generally improving both speed and accuracy of fundamental matrices and an iterative method to estimate the focal lengths from fundamental matrices.

In the repo for the paper [I got asked to contribute to PoseLib](https://github.com/kocurvik/robust_self_calibration/issues/1). I would gladly prepare a PR, but in one of the issues I found this relatively recent (Oct 2023) comment:

              @alexs7 the code is currently not in PoseLib. It's on my todo to add instrinsic refinement and once this is added I plan to add all of the corresponding solvers for unknown intrinsics.

_Originally posted by @vlarsson in https://github.com/PoseLib/PoseLib/issues/16#issuecomment-1746854692_

So I am not sure whether some code is already prepared. My question is therefore which of these would be welcome contributions. If any of them would be welcome, then I would appreciate some information on what testing or benchmarking is required for PRs.

vlarsson:
I saw your paper, nice work! It would be great if you could make a PR for this.

Regarding the comment, there is currently a WIP branch that refactors the camera models (https://github.com/PoseLib/PoseLib/tree/camera_models), and my original plan was to look into intrinsic+distortion parameter refinement after this, but I am not sure when I will find the time for this :) But even then, when this happens it will be easier if your code is integrated already.

vlarsson:
Regarding testing. If you want, you could also setup a PR for https://github.com/vlarsson/posebench that also tests the F estimators.

kocurvik:
> Regarding testing. If you want, you could also setup a PR for https://github.com/vlarsson/posebench that also tests the F estimators.

Thanks. I will look into it.

PR #102, opened 2024-07-26
Added decompositions for focal lengths from F

kocurvik:
Added decompositions as mentioned in #89 

There are two methods:
* `focals_from_fundamental` using [Bougnoux formula](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=710808&casa_token=SVXYLNsT1g0AAAAA:0Y9kjd_bwK5WicSdBA24gXfgj6tdkb3Jp5g6ogMVzFltczWB3l__wGGn7p-rm_lAmicEf-PaSk5B&tag=1)
*  `focals_from_fundamental_iterative` using [our recent iterative method](https://openaccess.thecvf.com/content/CVPR2024/papers/Kocur_Robust_Self-calibration_of_Focal_Lengths_from_the_Fundamental_Matrix_CVPR_2024_paper.pdf)

Both also have python bindings.

They both return two Camera structs. The iterative method also takes the prior camera info as the Camera struct and also outputs the number of iterations. The baseline method takes as input also the principal points of the two cameras.

I tested them on Phototourism with SP+SG matches:
together_phototourism_f

And the same setup on Aachen Day-Night v1.1:
together_aachen_f


PoseLib/decompositions.h
vlarsson, line 1: Maybe this should be under /misc/. At some point we should also add in homography factorization as well.

vlarsson (changes requested):
Looks great! Maybe we should move decompositions to /misc/ ?

PoseLib/decompositions.h
kocurvik: I made the change.

kocurvik:
I moved the files to misc and added a part about decompositions into the README, also bumped the version to 2.0.3.

vlarsson (approved):
Looks good to me!

PoseLib/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ set(SOURCES
     misc/essential.cc
     misc/re3q3.cc
     misc/colmap_models.cc
+        misc/decompositions.cc
```
pablospe, line 37: there are some extra spaces here, right?

kocurvik: Ah yes, my IDE always adds those. Forgot to remove them after moving the files to misc/

PR #104, opened 2024-08-04
Implementation of OpenCVFisheye camera model

vlarsson:
Add implementation of OpenCVFisheye camera model, as noted here https://github.com/colmap/glomap/issues/17#issuecomment-2265765607

Test consistency with pycolmap on randomly selected points for a couple of different cameras. The implementation is currently broken for wide FOV cameras (>180 degree) however. This will be addressed in https://github.com/PoseLib/PoseLib/tree/camera_models in the future.

### DIFF
--- a/PoseLib/misc/colmap_models.cc
+++ b/PoseLib/misc/colmap_models.cc
@@ -632,8 +632,9 @@ double opencv_fisheye_newton(const std::vector<double> &params, double rd, doubl
         if (std::abs(f) < UNDIST_TOL) {
             return std::abs(f);
         }
-        const double fp = (1.0 + 3.0 * theta2 * params[4] + 5.0 * theta4 * params[5] + 7.0 * theta6 * params[6] +
+        double fp = (1.0 + 3.0 * theta2 * params[4] + 5.0 * theta4 * params[5] + 7.0 * theta6 * params[6] +
                            9.0 * theta8 * params[7]);
+        fp += std::copysign(1e-10, fp);
         theta = theta - f / fp;
     }
     return std::abs(f);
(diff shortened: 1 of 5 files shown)


PoseLib/misc/colmap_models.cc
```diff
@@ -500,15 +555,127 @@ const std::vector<size_t> OpenCVCameraModel::principal_point_idx = {2, 3};
...
+        const double theta2 = theta * theta;
+        const double theta4 = theta2 * theta2;
+        const double theta6 = theta2 * theta4;
+        const double theta8 = theta2 * theta6;
+        f = theta * (1.0 + theta2 * params[4] + theta4 * params[5] + theta6 * params[6] + theta8 * params[7]) - rd;
+        if (std::abs(f) < UNDIST_TOL) {
+            return std::abs(f);
+        }
+        const double fp = (1.0 + 3.0 * theta2 * params[4] + 5.0 * theta4 * params[5] + 7.0 * theta6 * params[6] +
+                           9.0 * theta8 * params[7]);
```
ahojnnes, line 636: I guess there can be some edge case where fp becomes zero?

vlarsson: Hmm, yes that might be possible. I added a small shift away from zero. Did not seem to have any impact on the results.

vlarsson:
For some reason clang-format on my machine does not agree with the version on the CI.

pablospe:
> Edit: It seems to work reverting from clang-format version 18 to 16.

It is installed here:
https://github.com/PoseLib/PoseLib/blob/90f47ffce946569f00f410a528a6a7c4fce5fea2/.github/workflows/main.yml#L27

You can set the used version there if you want also in this PR.

lnexenl (commented):
Tested on my own dataset, seems to be no problem.

pablospe (approved):

Issue #103, opened 2024-08-01
Version tags


valgur:
Hi!

If it's not too much of a hassle, would it be possible to create Git tags or even GitHub releases for the current and any new versions?

I'm in the process of adding PoseLib as a package to ConanCenter (https://github.com/conan-io/conan-center-index/pull/24770) and I would like to use a real version string, such as the current `2.0.3`, as the package version, but the lack of corresponding Git tags makes it a bit ambiguous whether the current version in the repo is an upcoming or a finished release.

The tags also help with keeping track of newly added versions for package updates: https://qchateau.github.io/conan-center-bot/#/updatable

pablospe:
Something like this?
https://github.com/PoseLib/PoseLib/releases/tag/v2.0.3

@vlarsson: yes, it would be nice create releases of new versions. I was wondering what a good way would be to make this a bit more automatically. Currently the version is in this line:
https://github.com/PoseLib/PoseLib/blob/90f47ffce946569f00f410a528a6a7c4fce5fea2/CMakeLists.txt#L3

Probably it would be good to have a `version.txt` or something external that can be easily parsed? (also to be used in the python binding).

valgur:
@pablospe That's perfect. Thank you!

uilianries:
@pablospe Thank you for your quick action!! :clap:

pablospe:
One more tag:
https://github.com/PoseLib/PoseLib/releases/tag/v2.0.4

pablospe:
Hi @valgur. I was wondering if you could add a small section in the README, so people (including myself) know how to use this. That would be very useful!

uilianries:
@pablospe Hello! I'll open a PR with instructions!

PR #105, opened 2024-08-05
Bump pybind11 and version

vlarsson:
Bumps the pybind11 version to the latest. There seem to be some segfault issues that are resolved in this version. Also bump the version number to 2.0.4.

@pablospe @Parskatt it would be good to push new version to pypi as this should hopefully fix some of the crashes.

I've also added a flag to disable -march=native in the cmake which should be done for the wheels.


pablospe:
Update version here too:
https://github.com/PoseLib/PoseLib/blob/567a73b766632f3c03eb6fc422f4eaac0de9e214/setup.py#L77

pablospe:
The `PYTHON_PACKAGE` is only used here? It looks like `setup.py` and `.github/workflows/main.yml` is activating it, why? It looks like that there was some legacy code, it might be good to double check.

pablospe:
Following issue https://github.com/PoseLib/PoseLib/issues/103. Let's create a new release after this PR. The change log could potentially be moved (or simple copy it) to the release description.

pablospe:
> @pablospe @Parskatt it would be good to push new version to pypi as this should hopefully fix some of the crashes.

Yes, I was trying to follow up with this PR: https://github.com/PoseLib/PoseLib/pull/95.

pablospe (approved):

Parskatt (approved):
I think the least friction way is I just merge master into my fork and build the wheels from there. I might have time to do it tonight.

Parskatt:
Wheels should now be on pypi

PR #108, opened 2024-08-07
Added Table of Contents to the README.md

pablospe:
Now the README is quite big, and it is difficult to find a section. This should help to find the section from the top of the file.


pablospe:
well, we can simply use the tools from github...

image

PR #109, opened 2024-08-08
Added F + division model radial distortion solver with Tangent Sampson error

kocurvik:
So far the PR includes:
* [10 pt solver for F + two division model radial distortion params](https://openaccess.thecvf.com/content_iccv_2015/papers/Kukelova_Efficient_Solution_to_ICCV_2015_paper.pdf)
* standard 7 pt solver for F with sampling of rd params
* Camera struct for division model
* [Tangent Sampson error calculation for division model camera](https://openaccess.thecvf.com/content/ICCV2023/html/Terekhov_Tangent_Sampson_Error_Fast_Approximate_Two-view_Reprojection_Error_for_Central_ICCV_2023_paper.html)
* LO for F + two rd params with TSamp error - verified numerically
* FCamPair struct to hold the results of the estimation with python binding.

Some points to consider:
* I have created a struct `FCamPair` which holds one 3x3 F matrix and two camera structs. This is probably not ideal since F is for uncalibrated case and the camera structs only hold information on the radial distortion params. However, in terms of usage I found this to be the simplest solution.
* Both solvers assume that the distortion center is at (0, 0). This could be handled differently by having the user specify the distortion center.
* I added  `undistort_with_jac` method for the division model camera, and added runtime errors for all the other cameras. This returns the unnormalized undistorted point, but the Jacobian is for the normalized point. This is used for the TSamp computation.
* I also added `undistort` method in a similar manner. This is used for the sampling strategy. 
* Both methods are `2D->3D` since I assume that will be the case in the future.
* There is some leftover unused code for filtering for some reasonable rd params. This may improve perfomance, but with TSamp error it is not necessary. However, the bounds depends on the coordinate normalization so this could either be supplied by the used, made consistent somehow or completely removed.

Some benchmarking of these methods on real data will be made available soon. I can also later try add some benchmarking code for the 10 pt solver.

After the final structure and naming conventions will be decided I can add similar solvers for the case of the two cameras having the same rd params.


PoseLib/misc/colmap_models.cc
```diff
@@ -678,9 +749,58 @@ void OpenCVFisheyeCameraModel::unproject(const std::vector<double> &params, cons
...
+                                             Eigen::Vector3d *xu, Eigen::Matrix<double, 3, 2> *jac) {
+    Eigen::Matrix<double, 3, 2> dxudx;
+    dxudx << 1.0, 0.0, 0.0, 1.0, 2 * params[4] * x(0), 2 * params[4] * x(1);
+
+    Eigen::Matrix<double, 3, 1> h;
+    xu->topRows(2) = x;
+    xu->z() = 1 + params[4] * x.squaredNorm();
+    h = xu->normalized();
+
+    *jac = (Eigen::Matrix3d::Identity() - h * h.transpose()) * dxudx / xu->norm();
```
kocurvik, line 793: I will try to optimize this.

PoseLib/camera_pose.h
```diff
@@ -94,6 +94,22 @@ struct alignas(32) ImagePair {
 };
 
 typedef std::vector<ImagePair> ImagePairVector;
+
+struct alignas(32) FCam {
```
vlarsson, line 98: I agree the naming is not great here. Especially since FCam and FCamPair both refers to two-view geometry, so the analogy with Image and ImagePair is broken a bit. 

Maybe something like UncalibImagePair { Matrix3d, Camera, Camera } ? And then use it for both the shared and independent dist parameter case. But it's also a bit weird to call it UncalibImagePair if we have the cameras in the struct :) Maybe ProjectiveImagePair.

kocurvik: I changed this to ProjectiveImagePair

vlarsson:
Thinking more about it I think it would be best to avoid `undistort_with_jac`, mainly as it will be a pain to implement for all models.

I've been working a bit on the camera models refactor branch and there I've updated `project_with_jac` to also give jacobians w.r.t. the `params` vector. https://github.com/PoseLib/PoseLib/blob/camera_models/PoseLib/misc/camera_models.h#L50-L51
this is then used in the refinement here https://github.com/PoseLib/PoseLib/blob/camera_models/PoseLib/robust/optim/absolute.h#L77-L119

I think it might be better to setup the TSamp optimization in terms of `project_with_jac` using the closed form expression for pinv(J) instead. This should then be easier to allow to optimize (both F and cameras) with TSamp for any camera model. It might be a bit slower, but for constant camera params I guess it should be negligible as it can be pre-computed.

kocurvik:
I have checked the TSamp paper and I am not sure if I understand your comment correctly. I will use the notation from the paper to better summarize my thoughts.

I understand that using `proect_with_jac` makes sense for absolute pose since in that case you have a (normalized) 3D point $\vec{d}$ and then you can project it.

However, for relative pose problem with a backward model this seems to involve way more steps then necessary. So as I undestand it there are two ways of using TSamp:
* Current PR implementation: compute $\frac{\partial \pi^{-1}}{\partial \vec{h}}$ directly in `undistort_with_jac` without the need for the pseudoinverse calculation. It would be also possible to implement an `unproject_with_jac` or something like that.
* Using `project_with_jac`: This if I understand it correctly would require me to first unproject the point, normalize it, then call `project_with_jac`  (e.g. $\vec{p} = \pi(\vec{d})$ and $J = \frac{\partial \pi}{\partial \vec{d}}$) and then calculate $J^\dagger = \frac{\partial \pi^{-1}}{\partial \vec{h}}$ using the pseudoinverse of $J$ at $\vec{d}$.


Now, if all the Jacobians can be pre-computed I would agree that the second approach is way easier to implement as the unprojection code is already done for most (maybe all?) camera models. However, specifically for the problem of F + distortion params it is necessary to recompute the Jacobians for every model and this is also necessary for every LO step. Thus making the Jacobian calculation as efficient as possible seems to be very important to the overall performance.

My suggestion would be to maybe add something like `unproject_with_jac` which defaults to the second option (`unproject` -> `project_with_jac` -> `pinv`) if nothing faster (e.g. direct computation of jacobian without pseudoinverse) is available. This would still be slower than `undistort_with_jac` for the F + division model rd estimator since there would be some nops related to unprojection with $f = 1, \vec{c} = \vec{0}$, but that would not be as costly.

vlarsson:
You are right, we could do default implementations with (`unproject` -> `project_with_jac` -> `pinv`) for camera models where we dont have a specialized the implementation. I agree that this is the best solution.

I think only having `unproject` and not `undistort` is more clear, and I guess the difference in runtime would be negligible.

kocurvik:
I tried to merge your code from the `camera_models` branch and there is one thing that I am confused by. Maybe my understanding is incorrect.

I would expect the projection for the division model to perform these steps on an unprojected 3D point $\vec{x}_u$ to obtain the projected point $\vec{x}_p = distort(K \vec{x}_u)$.  This means that to obtain the unprojected point one would use $\vec{x}_u = K^{-1} undistort(\vec{x}_p)$. 

As I understand it the distorion and undistortion for the division model are both maps from image space to image space. I base this on the fact that in [the Fitzgibbon's paper](https://www.robots.ox.ac.uk/~vgg/publications/2001/Fitzgibbon01b/fitzgibbon01b.pdf) in section 4.2 he defines the epipolar constraint with $F$ on the undistorted points thus it seems to me that undistortion has to happen in the image space and not the unprojected space. However, reading the code in `camera_models` branch it seems to me that the order of these operations is reversed. If $f_x = f_y$ then this is only affects the scale of the distortion param. Otherwise it also makes the distortion elliptical w.r.t image coordinates.

vlarsson:
I think having it as `xd = K distort(xu)` is more natural. Otherwise you will either need to separate out the principal point from K or have the distortion compute r2 w.r.t. the principal point. It's also more consistent with respect to the other camera models which typically add K at the end.

kocurvik:
I have added the changes from `camera_models` branch. I did not do a git-based merge, but manually added the files. I don't know whether this would be an issue.

What was added from `camera_models`:
* all of `camera_models.cc` and `camera_models.h`
* tests for cameras
* small change to some `Camera` method calls in `jacobian_impl.h`
* modification of the main `CMakeLists.txt` - There are some differences, but I am not very experienced with CMake so I am not sure how much this affects other things.

What I added:
* `unproject_with_jac` for all cameras using the pseudoinverse
* custom `unproject_with_jac` for the division model - I checked and it is ~37% faster
* two test cases for the division model
* python bindings for Camera - all constructors + (un)project methods including jacobians and vectorized/single point versions

Currently there are not test cases for the `SphericalCameraModel`. I can either remove it from this PR or if someone has some reasonable params we can add a test case.

I will look into creating a benchmark for the added 10pt solver.

PoseLib/misc/camera_models.cc
```diff
@@ -0,0 +1,2170 @@
...
+        for (size_t idx : Model::principal_point_idx)                                                                  \
+            params.at(idx) *= scale;                                                                                   \
+        break;
+
+    switch (model_id) { SWITCH_CAMERA_MODELS }
+#undef SWITCH_CAMERA_MODEL_CASE
+}
+
+#define SWITCH_CAMERA_MODEL_CASE(Model)                                                                                \
+    void Model::unproject_with_jac(const std::vector<double> &params, const Eigen::Vector2d &xp, Eigen::Vector3d *x,   \
```
vlarsson, line 347: I guess we also want this to return jacobian w.r.t. distortion parameters?

` std::vector<Eigen::Matrix<double, 3, Eigen::Dynamic>> *jac_params`

Should be possible to derive based on the closed form expression of the psuedo-inverse. This is probably also be faster than `completeOrthogonalDecomposition` used here

kocurvik:
I rechecked the computation times for the custom implementation of unproject_with_jac for the division model and I previosly had an error there. In this commit it seems that the custom implementation is almost 60% faster than the default one.

vlarsson:
I did a quick merge between `camera_models` and `absolute_pose_unknown_clean` onto `dev` now. I removed the `estimate_absolute_pose_focal`, and merged it into `estimate_absolute_pose` as this https://github.com/PoseLib/PoseLib/blob/dev/PoseLib/robust.cc#L50-L58

kocurvik:
Should I change the base of this PR to the dev branch?

vlarsson:
Yeah, let's merge it there first and sort out any issues.

kocurvik:
I added the benchmark for the solver and merged with other changes in the dev branch. I will add tests for the jacobian calculation later.

kocurvik:
I added the tests. There was an issue with F flipping signs when stepping due to the Factorization. This seems to arise due to the fact that sometimes `FactorizedFundamentalMatrix(F).F() ~ -FactorizedFundamentalMatrix(F_step).F()` . This also results in all the residual signs being flipped even though the final sum of squares is the same.

I do not think this actually has any effect on the actual refinemet code, but it makes the computation of the finite jacobian used in the tests wrong (sometimes wrong sign, sometimes the error cancels out the forward and backward steps).

To remedy this I added the factorized F as  `ProjectiveImagePair` member. The issue is that when F gets assigned this is not coupled with the F and vice versa. This is not an issue now, but could lead to bugs in the future when someone assigns F and then the optimization starts from the old factorized F. I don't know how important is it to deal with this.

vlarsson:
Perhaps we should only keep the factorized version in `ProjectiveImagePair` and remove `Matrix3d`. Could expose it via something like
`Eigen::Matrix3d ProjectiveImagePair::F() { return FF.F() }; `
This might require a few more SVDs when we are assigning to FF though. I guess it is assigned in the minimal solver as well? Not sure how much of an impact this will have.

kocurvik:
I thought about it now and maybe we can just create `FactorizedProjectiveImagePair` which is only used during optimization. Similar to how `FactorizedFundamentalMatrix` is used only within the optim module.

vlarsson:
> I thought about it now and maybe we can just create `FactorizedProjectiveImagePair` which is only used during optimization. Similar to how `FactorizedFundamentalMatrix` is used only within the optim module.

Ah yes, this is probably the nicest solution.

vlarsson:
@kocurvik is this branch ready to merge now?

kocurvik:
I think it is almost ready.

The only thing I am unsure about is the minus sign in the `jac_param` calculation in the `unproject_with_jac` which I think should not be there, but the numerical tests indicate that it should. I tried checking for some errors that would lead to this, but did not find any so far.

vlarsson:
> I think it is almost ready.
> 
> The only thing I am unsure about is the minus sign in the `jac_param` calculation in the `unproject_with_jac` which I think should not be there, but the numerical tests indicate that it should. I tried checking for some errors that would lead to this, but did not find any so far.

Hmm, yes the minus sign is a bit surprising.

kocurvik:
@vlarsson My quick initial derivation of the formula by just using the chain rule was completely wrong and only by luck it required just the minus sign to be correct.

Now I have derived the expression from the fact that $unproj(x, p) = unproj(proj(X, p), p) = X$. If you take the partial derivative w.r.t. X of the last equation you get $\frac{\partial unproj}{\partial x}\frac{\partial proj}{\partial p} + \frac{\partial unproj}{\partial p} = 0$. Thus the minus sign. 

I also fixed the formatting. This PR can now be merged.

kocurvik:
I added the sampling and the 9pt solvers for cameras with shared radial distortion parameters. I also added jacobian tests and benchmark.

The benchmark looks like this now:
```
                    Solver    Solutions        Valid     GT found      Runtime
------------------------------------------------------------------------------
                       p3p      1.72803          100          100   340.699 ns
                  p3p_ding      1.72803          100       99.999   302.119 ns
                      gp3p        3.822          100        99.99   1.67507 us
                     gp4ps            1          100          100   2.07308 us
               gp4ps(Deg.)       2.2932          100        99.98   358.588 ns
                      p4pf       0.9999          100        99.19   2.75132 us
                    p3.5pf       4.4634      23.1057        99.86   16.9986 us
                      p5pf       2.2822        43.87          100   1.48329 us
                    p2p2pl        6.636      99.6986         99.6   25.6462 us
                      p6lp       4.4216          100          100   1.81627 us
               p5lp_radial      2.27858      99.9978       99.999   973.489 ns
                    p2p1ll            4      99.9975        99.99   1.49222 us
                    p1p2ll       4.4636          100          100    1.6408 us
                      p3ll       4.5886          100        99.95   1.78125 us
                      up2p            2          100          100   108.081 ns
                   up1p1ll            2          100          100   86.7726 ns
                     ugp2p            2          100          100   111.202 ns
                    ugp3ps            1          100          100   411.432 ns
                   up1p2pl       3.2774          100          100   408.882 ns
                     up4pl       4.5236       99.958        99.89   1.38131 us
                    ugp4pl       4.4442       99.928        99.78    1.4327 us
             RelUpright3pt       2.5128       99.992        99.99   374.125 ns
          GenRelUpright4pt       4.6264      99.8357        99.45   1.42691 us
                    Rel8pt            1          100          100   2.67907 us
           Rel8pt(100 pts)            1          100          100   6.69169 us
                    Rel5pt        2.253      98.5042        98.46   5.46871 us
         SharedFocalRel6pt       1.8449      98.0704        97.81   23.3073 us
                 RelRD10pt       6.0744      99.3102          100   14.4691 us
            RelSharedRD9pt       3.0308      11.0499        95.83   43.6136 us
       RelUprightPlanar2pt       1.8113          100          100   249.447 ns
       RelUprightPlanar3pt            1          100          100   334.737 ns
               GenRel5p1pt       1.1635      98.7881        47.88   5.34973 us
                 GenRel6pt        2.606      99.9616          100   727.153 us
             Homography4pt            1          100          100    487.71 ns
          Homography4pt(C)      0.99204          100       99.204   481.547 ns
```

The valid percentage looks quite low, but this is probably due to the nature of the solver which ignores the `det(F) = 0` constraint and it is not explicitly enforced in the solver as that is taken care of in the LO. Lowevering the tolerance to `1e-3` yields `73.8848` valid samples.

PR #111, opened 2024-09-03
Add FULL_OPENCV camera model

ahojnnes:
Used the following Python code to compute the Jacobian:
```
from sympy import symbols, diff

u, v, fx, fy, cx, cy, k1, k2, p1, p2, k3, k4, k5, k6 = symbols("u v fx fy cx cy k1 k2 p1 p2 k3 k4 k5 k6")

u2 = u * u
uv = u * v
v2 = v * v
r2 = u * u + v * v
r4 = r2 * r2
r6 = r4 * r2
alpha = (1.0 + k1 * r2 + k2 * r4 + k3 * r6) / (1.0 + k4 * r2 + k5 * r4 + k6 * r6)
xp0 = alpha * u + 2.0 * p1 * uv + p2 * (r2 + 2.0 * u2)
xp1 = alpha * v + 2.0 * p2 * uv + p1 * (r2 + 2.0 * v2)

print("J00")
print(diff(xp0, u))
print("J01")
print(diff(xp0, v))
print("J10")
print(diff(xp1, u))
print("J11")
print(diff(xp1, v))
```

### DIFF
--- a/PoseLib/misc/colmap_models.cc
+++ b/PoseLib/misc/colmap_models.cc
@@ -682,6 +682,102 @@ const std::vector<size_t> OpenCVFisheyeCameraModel::focal_idx = {0, 1};
 const std::vector<size_t> OpenCVFisheyeCameraModel::principal_point_idx = {2, 3};
 
 ///////////////////////////////////////////////////////////////////
+// FullOpenCV camera
+//   params = fx, fy, cx, cy, k1, k2, p1, p2
+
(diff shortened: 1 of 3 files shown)


vlarsson:
The FullOpenCV model is integrated in current dev branch here
https://github.com/PoseLib/PoseLib/blob/dev/PoseLib/misc/camera_models.cc#L1073-L1234
this should be merged into master once https://github.com/PoseLib/PoseLib/pull/109 has converged and is tested.

PR #115, opened 2024-10-10
Change default -march=native to be disabled

vlarsson:


pablospe:
Maybe include a description why of this change.

pablospe (approved):

pablospe:
Now that I thought about it longer, I believe being ON was a better default, and for distribution can be set OFF. At least this should be documented that users can change this for more performance on a specific machine (most of the time this will be compiled in a specific machine, reason why I think it was a better default). What was the reason of the change?

PR #116, opened 2024-10-22
Fix pybind11 calling overhead

vlarsson:
Playing around a bit with dense matches it seems that the pybind is quite slow. Adding this file (stolen from pycolmap) seem to improve runtime a bit. Without it there seems to be a 4-6ms overhead calling from python with 10k matches.

Maybe @sarlinpe or @Phil26AT have some more insight. In particular if anything needs to be adapted, or could be removed.


pybind/pybind11_extension.h
```diff
@@ -0,0 +1,125 @@
...
+
+  static handle cast(const std::vector<VectorType>& vec,
+                     return_value_policy /* policy */,
+                     handle h) {
+    Eigen::Map<const MatrixType> mat(
+        reinterpret_cast<const Scalar*>(vec.data()), vec.size(), Size);
+    return type_caster<Eigen::Map<const MatrixType>>::cast(
+        mat, return_value_policy::copy, h);
+  }
+};
```
sarlinpe, line 86: You only need this code block, which should indeed speed up all conversions to `std::vector<Eigen::Vector>`. I'd appreciate due credit being given since it is copied from COLMAP and I wrote it :)

vlarsson: Perfect, thanks! Of course, I'll add the proper attribution before merging it :)

pablospe (approved):

PR #121, opened 2024-11-23
Added H decomposition

kocurvik:
Added homography decomposition into pose using SVD.

Original code by @yaqding I just changed the call signature and added bindings.

If approved I will add info to README and bump version.


PoseLib/misc/decompositions.cc
```diff
@@ -1212,4 +1212,80 @@ std::tuple<Camera, Camera, int> focals_from_fundamental_iterative(const Eigen::M
...
+    }
+
+    Eigen::JacobiSVD<Eigen::MatrixXd> svd(HH, Eigen::ComputeFullV);
+    Eigen::Matrix3d H2 = HH / svd.singularValues()[1];
+
+    Eigen::Vector3d S2 = svd.singularValues();
+    Eigen::Matrix3d V2 = svd.matrixV();
+
+    if (abs(S2(0) - S2(2)) < 1.0e-6) {
+        poses->emplace_back(H2, Eigen::Vector3d(0.0, 0.0, 0.0));
```
vlarsson, line 1230: Should maybe also add a zero vector to normals here. 

Also this check could be made invariant w.r.t. the scale of HH.

PoseLib/misc/decompositions.h
```diff
@@ -19,6 +20,10 @@ focals_from_fundamental_iterative(const Eigen::Matrix3d &F, const Camera &camera
                                   const int &max_iters = 50,
                                   const Eigen::Vector4d &weights = Eigen::Vector4d(5.0e-4, 1.0, 5.0e-4, 1.0));
 
+// Estimate the camera motion from homography.
+// If you use H obtained using correspondences in image coordinates from two cameras you need to input K2^-1 * H * K1.
+void motion_from_homography(Eigen::Matrix3d HH, std::vector<CameraPose> *poses, std::vector<Eigen::Vector3d> *normals);
```
vlarsson, line 25: HH could be const ref unless we need to copy.

PoseLib/misc/decompositions.cc
```diff
@@ -1212,4 +1212,80 @@ std::tuple<Camera, Camera, int> focals_from_fundamental_iterative(const Eigen::M
...
+    Eigen::Vector3d t1 = (H2 - R1) * n1;
+
+    Eigen::Vector3d n2 = v2.cross(u2);
+
+    if (n2(2) < 0) {
+        n2 *= -1.0;
+    }
+    Eigen::Vector3d t2 = (H2 - R2) * n2;
+
+    poses->emplace_back(R1, t1);
```
vlarsson, line 1286: I guess there are four possible solutions here
(R1, t1, n1)
(R1, -t1, -n1)
(R2, t2, n2)
(R2, -t2, -n2)
So here we choose the solution pairs with n(2) > 0, but is it possible to decide that these are the correct ones without any correspondences?

vlarsson: Might be better to return all four otherwise.

PoseLib/misc/decompositions.cc
kocurvik: So maybe scale this w.r.t. the determinant?

PoseLib/misc/decompositions.h
kocurvik: This is done so we can enforce the determinant to be positive inside the function.

PoseLib/misc/decompositions.cc
vlarsson: Could divide by S2(0) as well.

kocurvik:
I forgot to comment. The requested changes have been committed. Let me know if this is ok so I can also add changes to README.

pablospe (approved):

PoseLib/misc/decompositions.cc
vlarsson: I think it's best to have `poses.size() == normals.size()` even in these corner cases. So  `normals->emplace_back(0,0,0)` is reasonable here as well I think.

vlarsson:
Looks good to me, thanks! Please fix the normals thing in the comment, and then update the README and we can merge.

kocurvik:
I added the normals output and description in README.

Should I also bump the version? I noticed there have been some other somewhat significant commits without version change since 2.0.4.

vlarsson (approved):
LGTM

PR #123, opened 2024-12-03
Release the GIL whenever possible to allow multithreading from python

theartful:
This enables multithreading from within python. Example experiment:
```python
import concurrent.futures
import time

import numpy as np
import poselib

num_tasks = 100

# Prepare task
camera = {
    "model": "SIMPLE_PINHOLE",
    "width": 1280,
    "height": 720,
    "params": [1.2 * 1280, 1280/2, 720/2]
}
points2D_1 = np.random.uniform(low=0.0, high=720, size=(100, 2))
points2D_2 = np.random.uniform(low=0.0, high=720, size=(100, 2))

now = time.time()

with concurrent.futures.ThreadPoolExecutor() as e:
    for _ in range(num_tasks):
        e.submit(poselib.estimate_relative_pose, points2D_1, points2D_2,
                 camera, camera)

then = time.time()

print(f"Done in {then - now} seconds")
```
Output on my machine before:
```
Done in 46.720627784729004 seconds
```
After:
```
Done in 2.9223525524139404 seconds
```

### DIFF
--- a/pybind/pyposelib.cc
+++ b/pybind/pyposelib.cc
@@ -30,6 +30,7 @@ py::dict BundleOptions_wrapper(py::dict overwrite) {
 }
 
 std::vector<CameraPose> p3p_wrapper(const std::vector<Eigen::Vector3d> &x, const std::vector<Eigen::Vector3d> &X) {
+    py::gil_scoped_release release;
     std::vector<CameraPose> output;
     p3p(x, X, &output);
     return output;
@@ -37,6 +38,7 @@ std::vector<CameraPose> p3p_wrapper(const std::vector<Eigen::Vector3d> &x, const
 
 std::vector<CameraPose> gp3p_wrapper(const std::vector<Eigen::Vector3d> &p, const std::vector<Eigen::Vector3d> &x,
                                      const std::vector<Eigen::Vector3d> &X) {
+    py::gil_scoped_release release;
     std::vector<CameraPose> output;
     gp3p(p, x, X, &output);
     return output;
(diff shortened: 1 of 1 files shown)


pablospe (approved):

pablospe:
@vlarsson Check this PR is you are ok with it.

vlarsson:
My understanding of the python GIL is limited. I don't know if there is any benefit in having some of these as callguards `m.def("foo", &foo, py::call_guard<py::gil_scoped_release>());` ? For the ones where the GIL is released for the entire function call. 

So am I correct in understanding that we need to have the GIL whenever we write to py::dict? Or why are we re-acquiring the GIL in some of these?

theartful:
@vlarsson as far as I know, whenever we deal with python objects we should hold the GIL. When writing this PR, I experimented with simply always releasing the GIL, but unsurprisingly I got crashes. Maybe writing an entirely new object (py::dict) can be ok, but I wouldn't risk it.

theartful:
`m.def("foo", &foo, py::call_guard<py::gil_scoped_release>());` Is probably cleaner. I will change the simple functions to be use that.

I will also experiment with not re-acquiring the GIL when writing `py::dict`s and get back to you.

theartful:
@vlarsson Holding the GIL while writing `py::dict` is necessary. I tried changing `estimate_relative_pose_wrapper` to release the GIL, and not re-acquire it back, and I reran the same experiment I shared in the PR, and I got segfault.

I changed the code to use `py::call_guard` whenever possible.

vlarsson:
Ah sorry, the recent PR #125 added some merge conflicts to this. Could you fix this? and then we can merge it.

theartful:
Should be ok now

PR #124, opened 2024-12-13
Wrapper for relative pose with non-upright gravity

B1ueber2y:


PoseLib/solvers/relpose_upright_3pt.h
```diff
@@ -41,6 +41,11 @@ namespace poselib {
 //   Y. Ding, J. Yang, V. Larsson, C. Olsson, K. Åström, Revisiting the P3P Problem, CVPR 2023
 int relpose_upright_3pt(const std::vector<Eigen::Vector3d> &x1, const std::vector<Eigen::Vector3d> &x2,
                         CameraPoseVector *output);
+
+// Wrapper for non-upright gravity (g_cam = R*g_world)
```
pablospe, line 45: Maybe you could clarify in the comment which one is the g in world? (g_cam1 or g_cam2). Such it is possible to understand without looking at the implementation.

pablospe (approved):

PoseLib/solvers/relpose_upright_3pt.h
B1ueber2y: Neither is g in world. They are the gravity in both camera frames. I think g in world is not needed in this relative pose setup as we only need to rotate the two frames to a frame with consistent "upright" vector, in this case [0 1 0]. I have updated the comment to avoid confusion.

vlarsson (approved):
LGTM

PR #129, opened 2025-01-07
Adding precision selection

seoj135:
This forked branch and PR purpose to add precision selection on PoseLib.
i.e, it allowes to use float precision instead of double.

Trivially, it has to tune parameters and make bigger thresholds/tolerances.

If you enabled `WITH_FLOAT` optionl on build, then you can use PoseLib float only.
And the results file will have midfix `f`
image
or `PoseLibfd.lib` on debug case

Also, this PR includes some improvements in benchmark, since there are overflow on runtime accumulation in debug builds.
And I'd like to save results automatically.

## MSVC Benchmark Results
[benchmark_results_double_0.000001.csv](https://github.com/user-attachments/files/18332217/benchmark_results_double_0.000001.csv)

[benchmark_results_float_0.000001.csv](https://github.com/user-attachments/files/18332218/benchmark_results_float_0.000001.csv)

[benchmark_results_float_0.001000.csv](https://github.com/user-attachments/files/18332220/benchmark_results_float_0.001000.csv)

## Clang Benchmark Results
[benchmark_results_double_0.000001.csv](https://github.com/user-attachments/files/18332221/benchmark_results_double_0.000001.csv)

[benchmark_results_float_0.000001.csv](https://github.com/user-attachments/files/18332223/benchmark_results_float_0.000001.csv)

[benchmark_results_float_0.001000.csv](https://github.com/user-attachments/files/18332224/benchmark_results_float_0.001000.csv)


PoseLib/real_t.h
```diff
@@ -0,0 +1,68 @@
...
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+// DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+// ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef POSELIB_REAL_TYPES_H_
```
pablospe, line 29: Perhaps something like this:
```
#ifndef POSELIB_REAL_T_H
#define POSELIB_REAL_T_H

#include <Eigen/Dense>

namespace poselib {

#ifdef USE_FLOAT
using Real = float;
#else
using Real = double;
#endif

// Define type aliases for Eigen types
using Vector2 = Eigen::Matrix<Real, 2, 1>;
using Vector3 = Eigen::Matrix<Real, 3, 1>;
using Matrix2x2 = Eigen::Matrix<Real, 2, 2>;
using Matrix3x3 = Eigen::Matrix<Real, 3, 3>;

} // namespace poselib

#endif // POSELIB_REAL_T_H
```

Such as one could do:
```
#include "PoseLib/real_t.h"

int main() {
    poselib::Vector2 v2;
    poselib::Vector3 v3;
    poselib::Matrix2x2 m2;
    poselib::Matrix3x3 m3;

    // You can now use v2, v3, m2, and m3 as Eigen types with the specified Real type (float or double)
}
```

Notice the use of `Real` (I feel `real_t` is harder to write), `USE_FLOAT`, and `poselib::XX` (instead of using Eigen::XX directly).

seoj135: okok got it

seoj135:
@pablospe I'd think 3.14 seems good to use as minimum, since it supports better `generator expression` syntax.
image
Surely if you guys accept it.

PoseLib/real_t.h
pablospe: These were only suggestions, let's see what @vlarsson thinks about it.

seoj135:
Note, 
* applied suggestions from @pablospe , but `real` instead of `Real`. Since `real` is primitive, I think lower-case name makes sense for it.
* separated type header to `real.h` and to `real_matrix.h` to improve build speed.
* Due to poor feature of cmake, my suggestion regarding cmake is abandoned.

pablospe:
> Due to poor feature of cmake, my suggestion regarding cmake is abandoned.

Could you expand on this? Not sure to understand about the poor feature of cmake. What did you want to achieve?


> * applied suggestions from @pablospe , but `real` instead of `Real`. Since `real` is primitive, I think lower-case name makes sense for it.

I feel using `Real` instead of `real` is a better option, for the following:

- **Consistency with Type Naming Conventions**:
  - It is a common convention to capitalize type names to distinguish them from primitive types and variables. This improves code readability and maintainability. See `types.h` and we are using all as VectorYYY, MatrixZZZ, etc.
  - Using `Real` aligns with this convention, making it clear that it is a custom type.

- **Readability and Maintainability**:
  - Capitalizing type names makes it easier to quickly identify and understand the code. It reduces cognitive load and helps in maintaining a consistent coding style.

- **Industry Standards and Best Practices**:
  - Refer to industry standards and best practices that recommend capitalizing type names. This can provide authoritative support for the change. For example, the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) recommends using capitalized names for types. And we might change to strictly follow Google C++ Style, see this [PR](https://github.com/PoseLib/PoseLib/pull/131/files#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2R1)


I would recommend changing it to `Real` (preferrable) or leaving it as `real_t`.

pybind/pyposelib.cc
```diff
@@ -935,21 +937,21 @@ PYBIND11_MODULE(poselib, m) {
         .def("prinicipal_point", &poselib::Camera::principal_point, "Returns the camera principal point.")
         .def("initialize_from_txt", &poselib::Camera::initialize_from_txt, "Initialize camera from a cameras.txt line")
         .def("project",
-             [](poselib::Camera &self, std::vector<Vector2> &xp) {
-                 std::vector<Vector2> x;
+             [](poselib::Camera &self, std::vector<poselib::Vector2> &xp) {
```
pablospe, line 940: One question here (and all the rest below). Shouldn't this be inside the `poselib` namespace?

seoj135: No, Since inter-language interface: `pybind, jni ... `, is worked on default namespace and C interface, each interface function wrapped with the macro `PYBIND11_MODULE` should be located on default namespace.

There is one way from me to resolve that boilerplate namespace calling here.
That is putting function-scope `using namespace poselib`. I guess it can solve the issue.

seoj135:
>> Due to poor feature of cmake, my suggestion regarding cmake is abandoned.
>
> Could you expand on this? Not sure to understand about the poor feature of cmake. What did you want to achieve?

Sorry for no defail about it.
Since cmake's generator expression doesn't work with `set()` command of cmake, i.e. only work with partial commands like `if()`.
It may need to increase cmake version too many. But I thought ppl prefer to keep lowest version of utilities, so I'd drop the change.

pablospe:
> It may need to increase cmake version too many. But I thought ppl prefer to keep lowest version of utilities, so I'd drop the change.

It would be ok to increase the cmake version, it is fine.

seoj135:
There are some misunderstanding regarding cmake from me. anyway cmake script is simplified.

PR #132, opened 2025-01-19
Remove gcc flags when using msvc for building pyposelib

theartful:
Building the python wheel on windows using msvc fails otherwise.


pablospe:
Are all the `target_compile_options` failing?
```
target_compile_options(pyposelib PRIVATE -march=native -Wall -Werror -fPIC -Wno-ignored-optimization-argument)
```

theartful:
@pablospe Windows doesn't have an equivalent to `march=native` as far as I know. Instead it has specific cpu feature flags. `-Wall` and `-Werror` has their equivalents: `/Wall` and `/WX`, see https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170
As for `-fPIC` there is no equivalent in MSVC.

So I think we're left with `/Wall` and `/WX`, and maybe we can check for CPU features and add `/arch:SSE2`, `/arch:AVX`, etc.

pablospe (approved):

pablospe:
@vlarsson Should we merge this?

vlarsson:
LGTM.

PR #127, opened 2024-12-18
Add overloaded wrapper methods for passing poselib::Camera instead of py::dict.

ghanning:
This pull request includes several changes to the `pybind/pyposelib.cc` file, primarily focusing on refactoring function signatures to use `Camera` objects directly instead of `py::dict` for camera parameters. Additionally, it introduces overloaded functions to maintain compatibility with the previous interface.

Refactoring for improved type safety:

* Modified `estimate_absolute_pose_wrapper`, `refine_absolute_pose_wrapper`, `estimate_absolute_pose_pnpl_wrapper`, and `refine_absolute_pose_pnpl_wrapper` to accept `Camera` objects directly instead of `py::dict` for camera parameters. [[1]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1L245-L250) [[2]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1R266-L276) [[3]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1R302-R315) [[4]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1L337-R363)
* Updated `estimate_generalized_absolute_pose_wrapper` and `refine_generalized_absolute_pose_wrapper` to use `Camera` objects instead of `py::dict` for camera parameters. [[1]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1L389-R424) [[2]](diffhunk://#diff-5afcda5201fab62f16e48f7fa083c666e5af05a67e3cd99096f5454806f763c1L418-R463)

Maintaining backward compatibility:

[...]

PR #135, opened 2025-04-07
Restore default arguments for `ransac_opt` and `bundle_opt`

ghanning:
... that were accidentally removed in #127.


vlarsson (approved):
LGTM

PR #145, opened 2025-06-18
SIMD Parallelization for Absolute Pose Estimation

xjiangan:
# Summary

This PR introduces SIMD parallelization to utilize modern CPU capabilities for faster absolute pose estimation in `poselib`.

Currently, poselib processes tasks such as RANSAC scoring and BA Jacobian accumulation in a purely sequential manner. These computations involve iterating over large sets of 2D/3D points, making them excellent candidates for parallelism.

# Key Changes

1. **Data Layout Adjustment**
   To enable SIMD-friendly memory access patterns, the storage order of point data has been modified. The original point data will be transposed at the beginning and each point’s dimensions are laid out contiguously in memory, improving vectorized reads and facilitating compiler SIMD optimizations.

2. **Compiler SIMD Auto-vectorization**
   Initial attempts were made to explore compiler auto-vectorization:

   * Clang was able to vectorize some loops when suitable compiler and optimization flags were provided.
   * GCC did not successfully vectorize these loops automatically.

3. **SIMD Parallelization (AVX Intrinsics)**
   A manual SIMD implementation has been provided for the **absolute pose estimation** process using AVX intrinsics. Benchmarks show this SIMD implementation achieves a **3× speed-up** over the previous scalar version. However, this hard-coded intrinsics requires CPU support for AVX instructions.

# Benchmarks

Performed using [posebench](https://github.com/vlarsson/posebench) with 
**gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0**
**Intel i7-14700K @ 5.6GHz** 

Finished running evaluation in 90.6 seconds (8 datasets)
Datasets: eth3d_130_dusmanu,7scenes_heads,7scenes_stairs,cambridge_landmarks_GreatCourt,cambridge_landmarks_ShopFacade,cambridge_landmarks_KingsCollege,cambridge_landmarks_StMarysChurch,cambridge_landmarks_OldHospital

| Method                  | AUC0  | AUC1  | AUC5  | med\_rot | med\_pos | avg\_rt   | med\_rt   |
| ----------------------- | ----- | ----- | ----- | -------- | -------- | --------- | --------- |
| PnP (poselib)           | 50.51 | 89.72 | 96.95 | 3.3E-01  | 0.1      | 7.1ms     | 6.4ms     |
| **PnP (poselib\_simd)** | 50.51 | 89.72 | 96.95 | 3.3E-01  | 0.1      | **2.3ms** | **2.1ms** |
| PnP (COLMAP)            | 50.77 | 89.72 | 97.02 | 3.4E-01  | 0.1      | 15.3ms    | 14.4ms    |

### DIFF
--- a/PoseLib/robust/estimators/absolute_pose.cc
+++ b/PoseLib/robust/estimators/absolute_pose.cc
@@ -64,7 +64,6 @@ void AbsolutePoseEstimator::refine_model(CameraPose *pose) const {
 
 #ifdef USE_SIMD_ABS_POSE
 
-
 void AbsolutePoseEstimatorSIMD::generate_models(std::vector<CameraPose> *models) {
     sampler.generate_sample(&sample);
     for (size_t k = 0; k < sample_sz; ++k) {
--- a/PoseLib/robust/utils.cc
+++ b/PoseLib/robust/utils.cc
@@ -188,7 +188,6 @@ void get_inliers_simd(const CameraPose &pose, const Eigen::MatrixX2d &x, const E
     }
 }
 
-
 #endif
 double compute_msac_score(const CameraPose &pose, const std::vector<Line2D> &lines2D,
                           const std::vector<Line3D> &lines3D, double sq_threshold, size_t *inlier_count) {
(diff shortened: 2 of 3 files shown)


Parskatt:
I think SIMD would be good to support, and performance benefits are pretty obvious from benchmarks.
Main concern is that it introduces a bunch of complexity into the codebase, also packaging is likely to get more complicated.
I might take a more detailed look at this later.

Issue #99, opened 2024-04-12
Camera pybind improvements


vlarsson:
For the poselib.Camera pybind 
https://github.com/PoseLib/PoseLib/blob/master/pybind/pyposelib.cc#L777-L785
some nice to have features could be

* Initialize from dict 
* Convert Camera to dict
* Interface for project/unproject

Parskatt:
Ah, should deal with this in https://github.com/PoseLib/PoseLib/pull/146

vlarsson:
In the dev branch there is now an `init_params()` that at least makes the camera valid.

https://github.com/PoseLib/PoseLib/blob/dev/PoseLib/misc/camera_models.h#L70-L78

This is called in the constructors.

Regarding initilizing `Camera()` as simple radial. Maybe. What are the use cases where you pass an empty `Camera()` where you want it to be simple radial? I guess, initializing it with zero distortion, unit focal makes it a calibrated camera.

vlarsson:
But I also agree that we should add some checks (e.g. number of params etc).

PR #146, opened 2025-07-28
Updated Python packaging + CI

Parskatt:
### Introduction
This PR introduces CI for publishing wheels to pypi (I've refrained from autopushing to pypi as it's immutable, so if you screw up you have to bump package version), and additionally comes with some bugfixes to the python bindings, and automatic python stub generation (courtesy of pycolmap). The CI is currently running over at [my fork](https://github.com/Parskatt/PoseLib), so any updates to this PR can be tested there first, before merging into master here.

### Changes
- Addition of cross-platform wheels for pypi through cibuildwheel in .github/workflows/build-pypi.yaml. 
- No longer build python package in .github/workflows/main.yaml, as python version now has some dependencies, keep main.yaml dep free.
- Unify `target_compile_options` across cmakelists as 

```cmake 
if(MSVC)
  target_compile_options(benchmark PRIVATE /bigobj /fp:fast)
else()
  if (MARCH_NATIVE)
    target_compile_options(benchmark PRIVATE
      -march=native -Wall -Werror -fPIC -Wno-sign-compare
      -Wno-unused-variable -ffast-math -Wno-ignored-optimization-argument)
  else()
    target_compile_options(benchmark PRIVATE
      -Wall -Werror -fPIC -Wno-sign-compare
      -Wno-unused-variable -ffast-math)
  endif()
  if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    target_compile_options(benchmark PRIVATE
      -Wno-maybe-uninitialized)
  endif()
endif()
```
- Some explicit `Eigen::Vector2d::Zero()` to satisfy picky compilers
- Follow [pycolmap](https://github.com/colmap/colmap/tree/main/python) and use _core for the python module, and move it into a folder `poselib` package folder, instead of just the raw .so.
- Add `pyproject.toml` to specify build-dependencies of the python package
- Add automatic stub generation, [following pycolmap](https://github.com/colmap/colmap/blob/main/python/generate_stubs.sh)
- Added constructor bindings for `poselib::Camera` (and also fixed segfault bug).

### TODO / Possible improvements
- Make equivalent of current `benchmark` for python module to be able to test for regressions etc. We might also consider running entire [`posebench`](https://github.com/vlarsson/posebench)
- Move away from dict-type wrappers for pybind and just bind the underlying classes/structs (e.g. RansacOptions etc). Currently it's a headache for me to remember the names, also easy to make mistakes.
- More explicit errors. Running something like relative pose estimation with a broken camera should result in an error, currently just returns 0 inliers.
- Update poselib.Camera (#99)

### Related PRs/Issues:
#72 
#73 
#99

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
 
 [project.optional-dependencies]
 test = [
-    "posebench@git+https://github.com/Parskatt/posebench.git ; python_version < '3.14'",
+    "posebench@git+https://github.com/PoseLib/posebench.git ; python_version < '3.14'",
     "pytest>=8.3.5",
 ]
 


setup.py
```diff
@@ -57,16 +61,20 @@ def build_extension(self, ext):
             build_args += ['--', '-j2']
 
         env = os.environ.copy()
+        
         env['CXXFLAGS'] = '{} -DVERSION_INFO=\\"{}\\"'.format(
-            env.get('CXXFLAGS', ''),
+            env.get('CXXFLAGS', ''), # + ' -fno-aligned-allocation' if platform.system() == "Darwin" else '',
```
pablospe, line 66: The comment can be implemented and then removed.

pablospe:
I was wondering what it is missing in order to merge, giving it is marked as Draft. Thanks for the work!

benchmark/CMakeLists.txt
```diff
@@ -21,7 +21,17 @@ install(
 if(MSVC)
   target_compile_options(benchmark PRIVATE /bigobj /fp:fast)
 else()
-  target_compile_options(benchmark PRIVATE
-    -march=native -Wall -Werror -Wno-sign-compare
-    -Wno-unused-variable -ffast-math)
+  if (MARCH_NATIVE)
```
vlarsson, line 24: On dev branch we have a cmake function for setting the compile options. Maybe we could backport this here instead of copy pasta

https://github.com/PoseLib/PoseLib/blob/dev/CMakeLists.txt#L17-L32

Parskatt: Should I merge in all of dev, or just this part? I agree the copypasting is bad.

pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+readme = "README.md"
+requires-python = ">=3.8"
+dependencies = [
+    "numpy",
+]
+
+[project.optional-dependencies]
+dev = [
+    "pybind11-stubgen",
+    "ruff",
```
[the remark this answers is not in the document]
Parskatt: We use ruff only in the build stage, following pycolmap, to format the stubs. (think of it as clang-format for python)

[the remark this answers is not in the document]
Parskatt: possibly we could remove the dev dependencies entirely for now, as we specify them anyway as have them anyway for build-deps.

benchmark/CMakeLists.txt
vlarsson: I would only do this part. I think dev is still partially broken and needs to be fixed before merging

pyproject.toml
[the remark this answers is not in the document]
vlarsson: Ah okay.

pybind/CMakeLists.txt
```diff
@@ -3,16 +3,42 @@
...
-    else()
-        target_compile_options(pyposelib PRIVATE -Wall -Werror -fPIC)
-    endif()
+set_target_properties(pyposelib PROPERTIES OUTPUT_NAME _core)
+set_compile_flags(pyposelib)
+
+# Install the compiled module
+install(TARGETS pyposelib LIBRARY DESTINATION . COMPONENT python)
+
+if(GENERATE_STUBS AND UNIX)
```
Parskatt, line 18: Should make this platform independent, currently we don't build any stubs for windows.

vlarsson:
> > Great work! Regarding the workflows. How does it work on github. In particular, if we later merge this branch into dev and get a `.github/workflows/main.yml` there as well. Will this affect the dev branch too? Or is it only ever master branch that can lead to packages being published to pypi?
> 
> This currently ever runs on master and on releases. Dev branch wouldn't be affected. It's pretty easy to change this behavior in the .yml if wanted.
> 
> > Or the plan is still to have manual publishing to pypi?
> 
> I'm uncomfortable with automatic publishing to pypi until we have set up some more proper tests for the python package.

Great. I agree that we can keep the publishing manual. Our releases are not frequent enough that it should be a problem :)

pablospe:
> I'm uncomfortable with automatic publishing to pypi until we have set up some more proper tests for the python package.

@Parskatt I’d like to better understand your concerns. Could you please explain how the testing process is related to whether we publish the package automatically or manually? Or, are you referring to testing the CI step that handles the publishing? My understanding is that, in both cases, the same package would be published with the same level of testing. 


> Great. I agree that we can keep the publishing manual. Our releases are not frequent enough that it should be a problem :)

@vlarsson I would argue that publishing the package from a fork is strange and give us less flexibility, not to mention that people need to trust the code it is in a fork, posing a security concern.

Parskatt:
@pablospe I meant that we do not automatically push to pypi, not that we should build on the fork. We can add a seperate push to pypi in the main repo later.

Parskatt:
Yeah, I can make a subset dataset which should be fine to dl many times.

pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+license = {file = "LICENSE"}
+readme = "README.md"
+requires-python = ">=3.8"
+dependencies = [
+    "numpy",
+]
+
+[project.optional-dependencies]
+test = [
+    "posebench @ git+https://github.com/Parskatt/posebench.git; python_version < \"3.14\"",
```
pablospe, line 21: As a reminder to change in the future: https://github.com/PoseLib/posebench

pablospe:
@Parskatt great, progress! Press the button "Ready for review" when you think this is ready and we go over; it looks already quite good.

Parskatt:
Thanks! I need to sort out a few things with python 3.14, but then should be ready. Don't want to feature-creep this PR too much.

Parskatt:
(Hopefully) final update:
- test command via pytest + posebench on a subset of the data that is not very heavy to run.

There are now a few new files.
`uv.lock` (dependency locking), `pyproject.toml`, the `pyposelib` folder, and the `tests` folder.

`uv.lock` could arguably be removed, and added to .gitignore. `uv` maintainers however recommend that it not be ignored. 
`pyproject.toml` should definitely remain. 
`pyposelib` needs to be somewhere for the package files. The naming can be arbitrary and we could move it later to a subfolder.
`tests` could possibly be renamed or be a subfolder of something else, it is however a pretty standard place for python packages to store tests.

I think these additions are fine, and pretty easy to revert later if we want to change the project structure.

pablospe:
yes, we should keep `uv.lock`. What I don't see it is where did you use it? I was expecting to find something like `uv pip sync`. Perhaps we need to add some documentation?

vlarsson:
> (Hopefully) final update:
> 
> * test command via pytest + posebench on a subset of the data that is not very heavy to run.
> 
> There are now a few new files. `uv.lock` (dependency locking), `pyproject.toml`, the `pyposelib` folder, and the `tests` folder.
> 
> `uv.lock` could arguably be removed, and added to .gitignore. `uv` maintainers however recommend that it not be ignored. `pyproject.toml` should definitely remain. `pyposelib` needs to be somewhere for the package files. The naming can be arbitrary and we could move it later to a subfolder. `tests` could possibly be renamed or be a subfolder of something else, it is however a pretty standard place for python packages to store tests.
> 
> I think these additions are fine, and pretty easy to revert later if we want to change the project structure.

Is this ready to merge now? Or do you have more changes planned?

Parskatt:
@vlarsson should be ready to merge.

@pablospe I used `uv` for building and dep management, but did not add any instructions, no. Basically just `uv sync` and `uv add` used. Should be noted that `uv` is still somewhat a pain to use with c-based projects as package is very often rebuilt, even for minor changes.

cmake/LibraryConfig.cmake
```diff
@@ -9,9 +9,6 @@ add_library(${LIBRARY_NAME}
 #   - PoseLib::PoseLib alias of PoseLib
 add_library(${PROJECT_NAME}::${LIBRARY_NAME} ALIAS ${LIBRARY_NAME})
 
-# C++17
-target_compile_features(${LIBRARY_NAME} PUBLIC cxx_std_17)
```
vlarsson, line 13: Why did we drop this? Or is it somewhere else?

pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+license = {file = "LICENSE"}
+readme = "README.md"
+requires-python = ">=3.8"
+dependencies = [
+    "numpy",
+]
+
+[project.optional-dependencies]
+test = [
+    "posebench@git+https://github.com/Parskatt/posebench.git ; python_version < '3.14'",
```
vlarsson, line 21: Update to PoseLib/posebench

pablospe:
@vlarsson Should we bump the version to `2.0.5`?
@Parskatt Do we need to only add the new tag, right? Or manually create a release? Or manually triggering the CI? How do we proceed here?

Parskatt:
CI will trigger for both release and push, so just changing the version in appropriate places will build wheels with correct tags. Then I'll push the wheels manually.

PR #147, opened 2025-07-29
Updated RANSAC stopping condition

alexs7:
@vlarsson If the current iterations reach a number over `state.dynamic_max_iter` it should also stop?


vlarsson:
It should always `run min_iterations` even if `dynamic_max_iter` is lower. So I believe the current code is correct?

PR #156, opened 2025-08-27
PoseBench as submodule, updates to ci test command, and some version increments for building wheels for pypi (#153)

### LINKED ISSUE #153: PyPI does not allow VCS dependencies
Pushing the 2.0.5 wheels to PyPI fails with the error:
                                                                                                     
```                                                                                                                                                                                                                          
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/                                               
         Can't have direct dependency: posebench@ git+https://github.com/PoseLib/posebench.git ; python_version <      
         "3.14" and extra == "test". See https://packaging.python.org/specifications/core-metadata for more            
         information.     
```                                                                                             

This seems to be due to the following rule: https://peps.python.org/pep-0440/#direct-references, which forbits direct references (like `posebench@ git+https://github.com/PoseLib/posebench.git`) for packaging.
It was not caught previously as it is allowed for building packages, just not distributing.

Another issue is that `posebench` is apparently already taken on PyPI: https://pypi.org/project/posebench/
Perhaps best to just integrate `posebench` fully into `poselib` as either a submodule or just copy-paste current commit repo into here?

@vlarsson, @pablospe opinions?

Parskatt:
Fixes #153 

As per https://github.com/PoseLib/PoseLib/issues/153 we need to avoid having direct dependencies, and seems easiest option for now is submoduling posebench. 

I also faced some issues with uploading wheels with twine (hence the bumps in setuptools and wheel, and the tool.setuptools). I'm not entirely sure that these changes are needed, so worth to revisit later. Anyway, now that things are set up we should probably run the pypi wheel pushing as trusted publishing (https://docs.pypi.org/trusted-publishers/) directly from the repo with the existing CI. I'll create an issue for this later.

As for tests we now simply pip install posebench directly in the CI test command, and proceed as before. Seems to work fine. Still skipping python 3.14 as h5py doesn't have wheels for it yet.


Parskatt:
And yeah, please ignore the 100 commits lol. Need to get off the fork or squash a bit...

PR #165, opened 2025-11-26
Remove Python 3.8 support + Run python wheel building CI for PRs.

B1ueber2y:
The wheel building is failing in the master branch for a while. Adding the checks to the CI workflow. Would be good to add these checks to the branch rulesets as well. 

Seems to also catch typos introduced in https://github.com/PoseLib/PoseLib/pull/152.

### DIFF
--- a/.github/workflows/build-pypi.yaml
+++ b/.github/workflows/build-pypi.yaml
@@ -21,18 +21,13 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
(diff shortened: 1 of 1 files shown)


kocurvik:
The critical bug preventing compilation was introduced when I accepted a suggestion https://github.com/PoseLib/PoseLib/pull/152/commits/6a78ec073e7cdd7564a2d652d5575abee4ac0fc9 from co-pilot fixing a typo, but only in one place. I probably compiled the PR without python bindings. Sorry for not catching it.

vlarsson (approved):
LGTM

PR #169, opened 2025-12-18
Add example estimator as comments for robust/ransac_impl.h.

B1ueber2y:
Edit: after discussion with @vlarsson I dropped inheritance and added an example class as comments in the header.


pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = [
 ]
 license = {file = "LICENSE"}
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
pablospe: nit: this is not very related to this PR

B1ueber2y: Yes this was missed from https://github.com/PoseLib/PoseLib/pull/165. I was too lazy to create a new PR for this so added it here.

pablospe (approved):

PR #170, opened 2025-12-18
Use #pragma once as include guard.

B1ueber2y:
Sister PR with https://github.com/colmap/colmap/pull/2007


pablospe (approved):

vlarsson:
Tested my local GCC installation and both with relative paths and with symlinks pragma once seemed to work. So maybe it is not an issue :)

PR #172, opened 2025-12-18
Split pyposelib.cc into multiple files according to the structures in C++.

B1ueber2y:
This file is long as crazy and cannot be well parsed by some of the AI agent.


B1ueber2y:
Dependent on upstream https://github.com/PoseLib/PoseLib/pull/171

PR #173, opened 2025-12-18
Add general impl for hybrid ransac.

B1ueber2y:
A general hybrid ransac impl with an example on point-line absolute pose estimation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,15 @@ find_package(Eigen3 REQUIRED)
 # Library sources
 add_subdirectory(${LIBRARY_FOLDER})
 
+# Sturm sequence solver max recursion depth
+# You should lower this value only if you need to run PoseLib in environments with low stack limit (<1MB).
+# Changing this may affect solver stability.
+set(MAX_STURM_RECURSION_DEPTH_LIMIT 300 CACHE STRING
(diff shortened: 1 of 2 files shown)


PoseLib/robust/hybrid_ransac_impl.h
```diff
@@ -0,0 +1,388 @@
...
+            bool found_valid = false;
+            for (size_t s = 0; s < num_solvers; ++s) {
+                if (prior_probs[s] > 0.0 && stats.num_iterations_per_solver[s] < state.dynamic_max_iter[s]) {
+                    found_valid = true;
+                    break;
+                }
+            }
+            if (!found_valid)
+                break;
+            continue; // Skip this solver, try again
```
vlarsson: I am not sure I understand the flow here. So if we select a solver that has reached max iters, we randomly select again? Or this cannot happen?

vlarsson (commented):
This looks good to me. Left some minor comments. 

Something that is a problem that we have to solve at some point is that the Camposeco approach does not really scale with the number of possible "solvers". So e.g. in multi-camera relative pose you might have 3-4 cameras on each rig, and the number of potential ways to select 6 correspondences spread out over these explodes. I remember @MikhailTerekhov looked into this at some point, but I don't think we ever got anything published.

PR #174, opened 2025-12-25
Implement NullCameraModel methods for projection and unprojection

cdcseacave:
This is an attempt to create a general camera type that can be used for example for relative pose estimation directly with unprojected coordinates, and so making it agnostic to the real camera type behind. Not only this should make poselib work with yet undefined camera models, like spherical camera, but also simplify the usage in a project that hosts multiple camera models and has already internal support for generating in a unified way (through virtual functions) the undistorted keypoint positions.


cdcseacave:
apart of the small extra computation, using PinholeCameraModel for general cameras is a bit misleading, ex setting PinholeCameraModel for estimating the relative pose for Spherical cameras...

PR #176, opened 2026-01-14
Add relpose_k2Fk1 solver from dev branch by @kocurvik 

ahojnnes:
All the credits to @kocurvik for the implementation.


kocurvik:
Note that with just these changes the solver won't work since the `DIVISION` model is not implemented in master and needs to be merged from dev as well. Since the type of camera model is declared via a string compilation won't fail.

PR #190, opened 2026-03-12
Migrate CI to uv

pablospe:
Migrate CI from pip to uv for reproducible builds via uv.lock.

- Replace setup-python + pip with astral-sh/setup-uv + uv sync --dev
- Add clang-format as a dev dependency in pyproject.toml
- Add .python-version (3.10)
- Bump actions/checkout to v4

Related: #189


pablospe:
Good catch — added `--no-install-project` so `uv sync` only installs clang-format without building the C++ extension. This keeps the step lightweight, matching the previous behavior.

PR #193, opened 2026-03-12
Clean test suite

vlarsson:
Updated tests to be more deterministic so we have fewer random failures in the CI (now it properly returns non-zero for failing tests).

Added support for suppressing output for passing tests.


B1ueber2y (commented):
I assume this is a half-finished PR since there are still some tests left not migrated to the new randomization tooling.

tests/test_rng.h
```diff
@@ -0,0 +1,111 @@
...
+
+inline uint64_t splitmix64(uint64_t x) {
+    x += 0x9e3779b97f4a7c15ULL;
+    x = (x ^ (x >> 30)) * 0xbf58476d1ce4e5b9ULL;
+    x = (x ^ (x >> 27)) * 0x94d049bb133111ebULL;
+    return x ^ (x >> 31);
+}
+
+inline uint64_t stable_string_hash(const std::string &text) {
+    uint64_t hash = 1469598103934665603ULL;
```
pablospe: FNV-1a 64-bit offset basis appears truncated — the standard value is `14695981039346656037ULL` (20 digits). The current constant `1469598103934665603ULL` (19 digits) is a different number entirely, which will produce a non-standard hash distribution. The prime `1099511628211ULL` on line 34 is correct.

tests/test_rng.h
```diff
@@ -0,0 +1,111 @@
...
+        return dist(engine_);
+    }
+
+    int uniform_int(int lower, int upper) {
+        std::uniform_int_distribution<int> dist(lower, upper);
+        return dist(engine_);
+    }
+
+    size_t uniform_index(size_t upper_exclusive) {
+        std::uniform_int_distribution<size_t> dist(0, upper_exclusive - 1);
```
pablospe: Latent issue: if `upper_exclusive == 0`, this wraps `size_t` to `SIZE_MAX`, creating a distribution over the full range. Not currently called with 0, but a defensive guard (e.g., `assert(upper_exclusive > 0)`) would prevent subtle misuse.

tests/optim_test_utils.h
```diff
@@ -142,4 +143,65 @@ inline bool check_valid_camera_projection(const Camera &camera, const Eigen::Vec
...
+    camera->rescale(1.0 / scale);
+    return scale;
+}
+
+inline double normalize_camera_points(std::vector<std::vector<Point2D>> &points, std::vector<Camera> *cameras) {
+    double scale = 1.0;
+    for (size_t camera_idx = 0; camera_idx < cameras->size(); ++camera_idx) {
+        scale = normalize_camera_points(points[camera_idx], &(*cameras)[camera_idx]);
+    }
+    return scale;
```
pablospe: This overload overwrites `scale` each iteration and returns only the last camera's scale. If the return value is ever used with heterogeneous multi-camera setups (different resolutions), the earlier scales are silently lost. Currently all call sites use the single-camera overload, so this is latent — but worth noting.

tests/test_rng.h
```diff
@@ -0,0 +1,111 @@
...
+    seed = combine_seed(seed, stable_string_hash(ctx.test_name));
+    seed = combine_seed(seed, stable_string_hash(case_name));
+    seed = combine_seed(seed, case_index + 1);
+    return seed;
+}
+
+inline unsigned int global_rand_seed() { return static_cast<unsigned int>(case_seed("__global__") & 0xffffffffULL); }
+
+inline unsigned int fixed_seed(const std::string &label, size_t case_index = 0) {
+    uint64_t seed = combine_seed(stable_string_hash(label), case_index + 1);
```
pablospe: `fixed_seed()` doesn't incorporate `suite_seed` or `repetition`, so stress runs (`--stress N`) won't vary seeds for callers using `fixed_seed`. I assume this is intentional (e.g., for reference poses that must remain constant)?

B1ueber2y (approved):

PR #196, opened 2026-03-16
Skip Python 3.14 in cibuildwheel

pablospe:
opencv-python lacks pre-built wheels for Python 3.14, causing macOS builds to fail on master.


pablospe:
But should we merge this to at least fix the CI in main? Until we plan what to do with posebench?Also, it might a good idea to merge posebench into this repo and use all the CI configuration from here, monorepo are easier to work with usually.

vlarsson:
I think we can close this PR now? Probably we just need to bump the posebench version in the CI?

PR #198, opened 2026-03-16
Fix stub generation in pip isolated builds and add ruff Python formatting

B1ueber2y:
* Fix generate_stubs.py to use sys.executable instead of passing the Python path as an argument. During pip's isolated builds, this ensures pybind11_stubgen and numpy are found correctly.
* Add scripts/ruff_format.sh for Python formatting (mirrors scripts/clang_format.sh)
* Add ruff configuration to pyproject.toml
* Pin ruff==0.15.0 in build-requires

### DIFF
--- a/pybind/generate_stubs.py
+++ b/pybind/generate_stubs.py
@@ -7,7 +7,6 @@ Adapted from generate_stubs.sh to work on Windows, macOS, and Linux.
 import sys
 import subprocess
 import re
-import shutil
 from pathlib import Path
 
 
@@ -66,17 +65,16 @@ def main():
     for file_path in files_to_process:
         process_stub_file(file_path)
 
-    # Format with ruff if available
-    if shutil.which("ruff"):
-        print("Formatting stubs with ruff...")
-        try:
-            subprocess.run(
-                ["ruff", "format"] + [str(f) for f in files_to_process], check=False
-            )
-        except Exception as e:
-            print(f"Warning: ruff formatting failed: {e}")
-    else:
-        print("ruff not found, skipping formatting")
+    # Format with ruff if available (use sys.executable -m to work in isolated builds)
+    print("Formatting stubs with ruff...")
+    try:
+        subprocess.run(
+            [sys.executable, "-m", "ruff", "format"]
+            + [str(f) for f in files_to_process],
+            check=False,
+        )
+    except Exception as e:
+        print(f"Warning: ruff formatting failed: {e}")
 
 
 def process_stub_file(file_path):
--- a/pybind/generate_stubs.py
+++ b/pybind/generate_stubs.py
@@ -5,7 +5,6 @@ Adapted from generate_stubs.sh to work on Windows, macOS, and Linux.
 """
 
 import sys
-import os
 import subprocess
 import re
 import shutil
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=68", "wheel>=0.38", "pybind11==3.0.1", "pybind11-stubgen", "ruff", "numpy"]
+requires = ["setuptools>=68", "wheel>=0.38", "pybind11==3.0.1", "pybind11-stubgen", "ruff==0.15.0", "numpy"]
```
pablospe: you might need to update the lock file too?

B1ueber2y: I dont personally use uv lock, so I honestly have no idea how to update it : )

B1ueber2y: Feel free to update directly on top of this branch.

pablospe: no worries. These are requirement for the build-system, they're installed in an isolated environment by pip/uv when building the wheel. They're not managed by uv.lock.

If you needed to update uv.lock (e.g., after changing project dependencies or dev groups):                                                                                                                       
                                                                                                                                                                                                                 
  `uv lock`
                                                                                                                                                                                                                 
  Or to also upgrade all dependencies to latest compatible versions:                                                                                                                                               
                                                                                                                                                                                                                   
  `uv lock --upgrade`

  But again, [build-system] requires is outside the scope of uv.lock. Apologies the confusion.

scripts/ruff_format.sh
```diff
@@ -0,0 +1,14 @@
...
+# Determine script absolute path
+SCRIPT_ABS_PATH=$(readlink -f ${BASH_SOURCE[0]})
+SCRIPT_ABS_PATH=$(dirname ${SCRIPT_ABS_PATH})
+
+# root folder
+ROOT="$(readlink -f ${SCRIPT_ABS_PATH}/..)"
+
+echo "Using $(ruff --version)"
+
+ruff check --fix ${ROOT}
```
pablospe: Maybe?
```suggestion
ruff check --fix "${ROOT}"
```

pyproject.toml
```diff
@@ -32,3 +32,10 @@ license-files = ["LICENSE"]
 dev = [
     "clang-format==21.1.8",
 ]
```
pablospe, line 35: I would add ruff as dev dependency, now that you added ruff_format.sh. Later do `uv sync --dev` (this installs ruff for you and update the lock)

```suggestion
dev = [
    "clang-format==21.1.8",
    "ruff==0.15.0",
]
```

pyproject.toml
pablospe: also pin `pybind11-stubgen` version?

pybind/generate_stubs.py
```diff
@@ -56,16 +61,18 @@ def main():
...
         sys.exit(1)
-    
+
     # Process each stub file
     for file_path in files_to_process:
         process_stub_file(file_path)
-    
+
     # Format with ruff if available
     if shutil.which("ruff"):
```
pablospe: `shutil.which("ruff")` might not find ruff in isolated builds (same issue this PR fixes for pybind11_stubgen)?

pybind/generate_stubs.py
```diff
@@ -56,16 +61,18 @@ def main():
...
         print("Formatting stubs with ruff...")
         try:
-            subprocess.run(["ruff", "format"] + [str(f) for f in files_to_process], check=False)
+            subprocess.run(
```
pablospe: `subprocess.run([sys.executable, "-m", "ruff", "format"] + [str(f) for f in files_to_process], check=False)`
                                                                                                                                                                                                                      
This runs ruff as a Python module (python -m ruff) instead of looking for it on PATH. Same pattern as running python -m pytest instead of pytest.

pybind/generate_stubs.py
B1ueber2y: Good catch. Thanks! This is fixed.

vlarsson (approved):
LGTM

PR #201, opened 2026-03-18
Fix the python executable for stubgen to be consistent.

B1ueber2y:
Without this fix, the python executable for stubgen was still confused by find_package(Python).


vlarsson (approved):
LGTM

pablospe (approved):

Issue #195, opened 2026-03-13
Radial distortion relative pose estimators with focal length estimation


kocurvik:
Congrats on the dev merge! Now that it is merged I have a suggestion for new features that I could implement.

Recently, we published a [paper dealing with radial distortion in relative pose estimation](https://rdcu.be/e73WK). This was a follow-up work on the paper after which PR #109 was made. The difference in the estimators in the follow-up work is that F gets decomposed before scoring (in case of equal cameras the sampling strategy uses the 6-point solver instead of 7-point). Then we optimize directly the pose, focal, and distortion instead of F + distortion. These estimators do not require new solvers, but just decomoposition and optimizers. I have [a fork with such estimators](https://github.com/kocurvik/PoseLib/tree/rdnet). It would need adapting a bit, but now with the `dev` merged I think it should be easy to implement.

The benefits would be that instead of `ProjectiveImagePair` we would return `ImagePair` so you would be guaranteed to get valid pose + intrinsics. I think that this is generally more useful, but may be unnecessary for things like global SfM.

We could also use the [recalibrator](https://github.com/PoseLib/PoseLib/blob/93f361b41f571a728805131a699adb64ad976c6c/PoseLib/robust/recalibrator.cc#L2) to output any of the implemented camera models not just the `SIMPLE_DIVISION` model. I would probably leave this for the final optimization step.

@vlarsson let me know whether I should make a PR with this and which variants would make sense to include.

PR #205, opened 2026-04-06
Estimator for relative pose with two unknown focal lenghts

kocurvik:
This PR adds an estimator with a refiner for the case of relative pose with two unknown focal lengths. The estimator uses the 7pt algorithm. Each F is decomposed into focals using the Bougnoux formula. If valid focals are obtained the pose is estimated from E considering the cheirality of inliers. The scoring uses Sampson of F while the refinement optimizes the pose and the focals directly.

In practice this is better than estimating F and decomposing it later as decomposition from a single F is more likely to fail. Still this does not work so well due to the common degeneracies.

Below you can compare the results for posebench (subset = True):
```
Finished running evaluation in 53.3 seconds (22 datasets)
Datasets: fisheye_grossmunster_4342,fisheye_kirchenge_2731,megadepth1500_sift,megadepth1500_spsg,megadepth1500_splg,megadepth1500_roma,megadepth1500_dkm,megadepth1500_aspanformer,scannet1500_sift,scannet1500_spsg,scannet1500_roma,scannet1500_dkm,scannet1500_aspanformer,imc_british_museum,imc_london_bridge,imc_piazza_san_marco,imc_florence_cathedral_side,imc_milan_cathedral,imc_sagrada_familia,imc_lincoln_memorial_statue,imc_mount_rushmore,imc_st_pauls_cathedral

                     AUC5   AUC10   AUC20   avg_rt   med_rt 
E (poselib)         51.89   65.63   76.33   29.6ms   27.9ms 
E (poselib,TS)      51.46   65.26   76.14   90.4ms   76.5ms 
F (poselib)         33.61   47.30   60.42   28.8ms   25.9ms 
F (poselib,decomp)  17.14   27.70   36.47   28.9ms   25.8ms 
F (poselib,vf)      19.54   32.07   44.86   23.8ms   21.5ms 
E (COLMAP)          45.47   61.43   73.90   20.0ms   17.0ms 
F (COLMAP)          31.24   42.84   56.37    7.3ms    5.5ms
```

The `decomp` version estimates F and decomposes it after the final model is provided. The `vf` version is the current PR. I imlemented this in posebench. Will add posebench PR soon.

### DIFF
--- a/PoseLib/robust.h
+++ b/PoseLib/robust.h
@@ -40,13 +40,13 @@
 namespace poselib {
 
 // Estimates absolute pose using LO-RANSAC followed by non-linear refinement
-// Threshold for reprojection error is set by RansacOptions.max_reproj_error
-// If ransac_opt.estimate_focal, the camera in image.camera should contain correct principal point
+// Threshold for reprojection error is set by AbsolutePoseOptions.max_error
+// If opt.estimate_focal_length, the camera in image.camera should contain correct principal point
 RansacStats estimate_absolute_pose(const std::vector<Point2D> &points2D, const std::vector<Point3D> &points3D,
                                    AbsolutePoseOptions opt, Image *image, std::vector<char> *inliers);
 
 // Estimates generalized absolute pose using LO-RANSAC followed by non-linear refinement
-// Threshold for reprojection error is set by RansacOptions.max_reproj_error
+// Threshold for reprojection error is set by AbsolutePoseOptions.max_error
 RansacStats estimate_generalized_absolute_pose(const std::vector<std::vector<Point2D>> &points2D,
                                                const std::vector<std::vector<Point3D>> &points3D,
                                                const std::vector<CameraPose> &camera_ext,
@@ -56,21 +56,22 @@ RansacStats estimate_generalized_absolute_pose(const std::vector<std::vector<Poi
 // Estimates absolute pose using LO-RANSAC followed by non-linear refinement
 // using both 2D-3D point and line matches
 // Note that line segments are described by their endpoints
-// Threshold for point reprojection error is set by RansacOptions.max_reproj_error
-// and for lines the threshold is set by RansacOptions.max_epipolar_error
+// Threshold for point reprojection error is set by AbsolutePoseOptions.max_errors[0]
+// and for lines the threshold is set by AbsolutePoseOptions.max_errors[1]
+// (If max_errors is empty, AbsolutePoseOptions.max_error is used for both)
 RansacStats estimate_absolute_pose_pnpl(const std::vector<Point2D> &points2D, const std::vector<Point3D> &points3D,
                                         const std::vector<Line2D> &line2D, const std::vector<Line3D> &line3D,
                                         const Camera &camera, const AbsolutePoseOptions &opt, CameraPose *pose,
                                         std::vector<char> *inliers_points, std::vector<char> *inliers_lines);
 
 // Estimates relative pose using LO-RANSAC followed by non-linear refinement
-// Threshold for Sampson error is set by RansacOptions.max_epipolar_error
+// Threshold for Sampson error is set by RelativePoseOptions.max_error
 RansacStats estimate_relative_pose(const std::vector<Point2D> &points2D_1, const std::vector<Point2D> &points2D_2,
                                    const Camera &camera1, const Camera &camera2, const RelativePoseOptions &opt,
                                    CameraPose *relative_pose, std::vector<char> *inliers);
 
 // Estimates relative geometry from using points and estimated depth using LO-RANSAC followed by non-linear refinement
-// Threshold for Sampson error is set by RansacOptions.max_epipolar_error
+// Threshold for reprojection and epipolar errors are set by MonoDepthRelativePoseOptions.max_errors
 // MonoDepth relative pose estimation with known calibration
 // Uses hybrid scoring with reprojection and epipolar errors
 RansacStats estimate_monodepth_relative_pose(const std::vector<Point2D> &points2D_1,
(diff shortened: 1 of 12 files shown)


Parskatt:
@kocurvik what does COLMAP currently run for initial pairs with no prior? Could this be upstreamed for the incremental version? I think poselib is gradually replacing their estimators? @B1ueber2y maybe you know more?

B1ueber2y:
COLMAP does not estimate focal lengths from two-view geometry until now. If no focal length is available (from prior or EXIF), colmap assumes 1.2 * max(w, h) and use this to transform F into E. From my understanding this is in practice good for typical consumer cameras. The focal lengths in colmap is only updated through pnpf and bundle adjustment. 

An idea to study this is to set up the 1.2x heuristics in posebench as well, but this would highly depend on the focal lengths in the dataset. Its also unclear how to use the feature from this PR as a fallback because we would need a criteria to tell whether the 1.2x heuristics is bad in the init.

Parskatt:
Ive had a lot of colmap failures due to the initial pair decomp failing. Would be nice if poselib features could propagate up there.
I guess the 1.2x is often ok after bundle for standard (no zoom etc) cameras. But the initial pair selection is really a pain point. I guess glomap handles it a bit more gracefully due to simultaneous solve.

kocurvik:
@Parskatt In my experience EXIF is generally quite accurate most of the time. I do not think that a point-based two-view solver can beat it over general distributions of images from common devices. There are some failure cases, but not many. Reverting to 1.2 * width heuristic is not that much worse compared to the alternatives - see Fig. 4 of [our CVPR 24' paper](https://arxiv.org/pdf/2311.16304). For that paper we also tried collecting a dataset with several phone cameras and their focals were all so close to the prior that nothing could beat it so we did not even include it. 

Though, if you expect that the cameras will be far from the prior and EXIF won't work, then this estimator could help. This estimator may be better than F + decomposition from our CVPR paper because you do not need priors and can directly estimate on all points and it is faster (I might benchmark this). Still the issue is that there is a generic degeneracy when pricipal axes intersect or are coplanar and this happens very often. Like any possible approach, this estimator suffers from this both during model generation (you get bad/imaginary/negative focals) and also during refinement where due to the instability around the degenerate config you essentially just select model which best fits the noise. You can avoid this if you have more than two views. Using some network for calibration from a single image can sidestep this issue entirely and imo this is the best approach if your data is close to their training distributions.

kocurvik:
@Parskatt I did the benchmarking here: https://github.com/PoseLib/posebench/pull/12#issuecomment-4204881174 you can see that the estimator from this PR beats decomposing F after RANSAC. The iterative decomposition is worse compared to the direct closed-form method, but that could be due to using GT principal point instead of center of image for most of the data since the image dimensions are not available.

Parskatt:
@kocurvik nice! To me the gains look significant and would be a nice fallback when exif is missing (without requiring to run a mono model). But that is besides the point. I'll have a look at the code when I get the chance, although @vlarsson might be a better judge.

kocurvik:
@Parskatt @B1ueber2y 

I added one more version to posebench (did not include it in PR). This time E solver is used with the colmap assumption that `focal = 1.2 * max(width, height)` or `2.4 * max(pp)` if not available.


The results are like this `(subset=False, subsample=10)`:
```
                               AUC5   AUC10   AUC20   avg_rt   med_rt 
E (poselib)                   51.73   65.46   76.15   28.1ms   26.1ms 
E (poselib,TS)                51.99   65.63   76.19   86.8ms   74.6ms 
E (poselib,colmap prior)       2.71    8.08   18.70   37.2ms   32.3ms 
F (poselib)                   31.21   44.06   56.22   25.6ms   22.3ms 
F (poselib,decomp)            15.94   26.08   35.32   25.6ms   22.2ms 
F (poselib,decomp iterative)  15.30   25.44   35.69   25.9ms   22.7ms 
F (poselib,decomp in RANSAC)  18.09   30.87   44.05   23.1ms   21.3ms 
E (COLMAP)                    45.74   61.08   73.44   19.7ms   17.2ms 
F (COLMAP)                    26.98   39.90   52.78    7.0ms    5.2ms 
```

This includes some fisheye scenes, but those could be addressed by sampling (#195) and for F solvers the distortion is not taken into account anyways. 

The colmap prior + E does not seem to work too well.

B1ueber2y:
Thank you! However, I think what colmap does in this case is to estimate F and use prior for converting to E, rather than use prior to estimate E. I assume this could make a difference.

kocurvik:
> I think what colmap does in this case is to estimate F and use prior for converting to E,

@B1ueber2y I tried this with the following result:
```
F (poselib,colmap prior + cheirality)         3.65   10.48   22.34   25.8ms   22.4ms 
F (poselib,colmap prior, all poses checked)   4.78   14.75   32.97   25.4ms   22.0ms 
```
Still seems significantly worse than using decomposition. The version with all poses does not care for cheirality and just takes best error out of all 4 possible poses.

B1ueber2y:
Interesting. Thank you! In this case, we should indeed upstream this as a fallback.

PoseLib/robust/bundle.h
```diff
@@ -110,9 +110,9 @@ BundleStats refine_shared_focal_relpose(const std::vector<Point2D> &x1, const st
                                         const std::vector<double> &weights = std::vector<double>());
 
 // TODO: Relative pose with two different unknown focals refinement not yet implemented in dev's refiner architecture
```
vlarsson: This PR fixes the TODO comment?

PoseLib/robust.h
```diff
@@ -86,6 +86,13 @@ RansacStats estimate_shared_focal_relative_pose(const std::vector<Point2D> &poin
                                                 const RelativePoseOptions &opt, ImagePair *image_pair,
                                                 std::vector<char> *inliers);
 
+// Estimates relative pose with two different unknown focal lengths using LO-RANSAC followed by non-linear refinement
+// Threshold for Sampson error is set by RansacOptions.max_epipolar_error
```
vlarsson: Maybe update comment here as well. RansacOptions.max_epipolar_error is no longer used

vlarsson: This also seems to be the case for some of the other ones. My fault for not updating this 😅 

Another minor nit here is that the interface is a bit inconsistent w.r.t. the principal point. For example `estimate_shared_focal_monodepth_relative_pose` just below does not take `pp` as input but assumes it's zero. Probably better to have the `pp` as an argument, less risk that a user incorrectly calls the function.

vlarsson: Could also note that we always output SIMPLE_PINHOLE cameras here.

vlarsson (approved):
Looks great! Thanks for putting this together. I mainly found some things related to the comments, but this seems to be a more widespread issue which we can fix in another PR.

kocurvik:
I modified the comments based on the used struct. I also updated the README to have an actual example. The monodepth estimators now take pp in args.

vlarsson (approved):
Finally got around to this. Sorry for the delay

PR #184, opened 2026-02-11
Merge dev branch into master

vlarsson:
WIP merge of the dev branch into master.

PR #206, opened 2026-04-22
feat(robust): add bearing-vector pose estimation API

cdcseacave:
## Overview

This PR adds a unified bearing-vector pose estimation API to PoseLib, enabling robust pose estimation for **any central camera model** (pinhole, spherical/equirectangular, fisheye, etc.) directly from 3D unit bearing vectors instead of 2D pixel coordinates.

## Motivation

With the addition of the spherical camera model to PoseLib (PR #184), there's a need for pose estimators that work natively with 3D bearings rather than requiring camera-specific unprojection to pixels. The bearing-vector approach:

- **Unifies camera handling**: Single solver path for pinhole, fisheye, spherical, and any future central camera model
- **Enables omnidirectional photogrammetry**: 360° cameras, panoramic imaging, and spherical image collections
- **Improves numerical stability**: Sphere-native error metrics (chord-distance, Sampson-on-the-sphere) avoid camera parameter singularities
- **Supports back-hemisphere reconstructions**: Essential for spherical cameras where valid points can have camera-space z < 0

### New Functions

**`estimate_absolute_pose_bearings()`**
- Absolute pose estimation from 3D unit bearing vectors and 3D world points
- LO-RANSAC + non-linear refinement pipeline
- Chord-distance scoring on unit sphere (no cheirality constraint)
- Full back-hemisphere support
- Use case: Camera localization from omnidirectional images

**`estimate_relative_pose_bearings()`**
- Relative pose from bearing correspondence pairs (e.g., two spherical images)
- Sampson-on-the-sphere error metric
- Bearing-native cheirality check (works for back-hemisphere points)
- Configurable angular error threshold via `RelativePoseOptions::SetMaxErrorFromAngle()`
- Use case: Two-view geometry from 360° video frames or panoramic image pairs

### Key Design Decisions

1. **Sphere-native error metrics**: 
   - Chord distance: `2 * sin(angle/2)` for absolute pose
   - Sampson-on-sphere: Generalizes 2D Sampson error to bearing space

2. **Back-hemisphere cheirality**: 
   - Triangulation via midpoint parameters (not z-sign check)
   - Enables reconstruction from spherical cameras where features can be observed from back-hemisphere

3. **API consistency**: 
   - Mirror pixel-based `estimate_absolute_pose()` and `estimate_relative_pose()` signatures
   - Options/output structures reused (AbsolutePoseOptions, RelativePoseOptions, CameraPose)

4. **Backward compatibility**: 
   - Pinhole bearings (b.z == 1) reduce to existing pixel-based solvers
   - No changes to existing pose estimators or camera models

## Testing & Integration

- **Compilation verified**: Builds cleanly on macOS Clang 17, no warnings
- **Integration tested**: Tested with OpenMVS spherical photogrammetry pipeline for dense reconstruction and mesh generation
- **Backward compatibility**: Existing pinhole and fisheye workflows unaffected

## Example Usage

```cpp
#include <PoseLib/robust.h>

// Convert spherical image observations to bearings
std::vector<Point3D> bearings;
std::vector<Point3D> world_points;
// ... populate from spherical image correspondences ...

// Estimate absolute pose
AbsolutePoseOptions opts;
opts.max_error = 0.05;  // chord-distance threshold
CameraPose pose;
std::vector<char> inliers;

RansacStats stats = poselib::estimate_absolute_pose_bearings(
    bearings, world_points, opts, &pose, &inliers);

// Use pose for localization
```

## Related

- Builds on: PR #184 (spherical camera model support)
- Used by: OpenMVS spherical photogrammetry pipeline
- Addresses: GitHub issue discussions on omnidirectional camera support

## Checklist

- [x] Compilation verified (no errors/warnings)
- [x] Backward compatibility maintained
- [x] API documentation in header comments
- [x] Integration tested with OpenMVS spherical workflows
- [x] Unit tests (can be added in follow-up PR)
- [x] Benchmark / performance characterization (for future iteration)

---

**Author's Note**: This contribution enables a key use case for spherical/omnidirectional photogrammetry. Happy to add unit tests, benchmarks, or make adjustments based on maintainer feedback.

---

## Benchmark Results — pinhole↔bearing parity (added 2026-05-05)

Standalone benchmark `tests/bearing_parity_bench` runs 100 randomized noisy pinhole scenes (focal = 800 px, Gaussian noise σ = 0.5 px) through both the pixel-based estimators (`estimate_absolute_pose`, `estimate_relative_pose`) and the bearing-based estimators (`estimate_*_pose_bearings`) on the same observations. Built locally on macOS / Apple Clang 17:

### Absolute pose (PnP, N=100)

| Path    | Median rotation err | Median translation err | Mean runtime |
|---------|--------------------:|-----------------------:|-------------:|
| Pinhole | 2.65 × 10⁻⁴ rad     | 1.73 × 10⁻³            | 1.14 ms      |
| Bearing | 2.91 × 10⁻⁴ rad     | 1.95 × 10⁻³            | 1.46 ms      |

### Relative pose (5pt, N=80)

| Path    | Median rotation err | Median translation-angle err | Mean runtime |
|---------|--------------------:|-----------------------------:|-------------:|
| Pinhole | 1.01 × 10⁻³ rad     | 4.10 × 10⁻³ rad              | 5.42 ms      |
| Bearing | 9.27 × 10⁻⁴ rad     | 3.56 × 10⁻³ rad              | 5.38 ms      |

The bearing path matches the pinhole path within ~10 % of pose error and is essentially the same speed for relative pose; the absolute-pose bearing path is ~30 % slower because the chord-distance Jacobian is denser than the (x, y)-pinhole Jacobian. Crucially, both paths recover the same pose within the noise floor — i.e. the bearing implementation is geometrically correct on pinhole inputs.

The in-tree unit tests (`test_pinhole_bearing_parity_absolute` / `_relative`, run as part of the standard `make run_tests`) additionally assert noise-free parity to ≤ 1 × 10⁻⁵ rad rotation / 1 × 10⁻⁵ translation for absolute pose and ≤ 1 × 10⁻⁴ rad rotation / cosine ≥ 1 − 1 × 10⁻⁶ for relative pose. All 87 unit tests pass.

### DIFF
--- a/PoseLib/robust/bundle.h
+++ b/PoseLib/robust/bundle.h
@@ -50,10 +50,18 @@ BundleStats bundle_adjust(const std::vector<Point2D> &x, const std::vector<Point
 
 // Absolute pose refinement for any central camera model (pinhole, spherical,
 // fisheye, ...) using 3D unit bearing vectors instead of 2D normalized pixels.
-// Minimizes the 3D residual r = normalize(R*X + t) - b_obs on the unit sphere.
-// No cheirality check — works for the full sphere (back-hemisphere points are
-// valid). For pinhole bearings this is algebraically equivalent to the 2D
-// bundle_adjust above.
+// Minimizes the 3D residual r = normalize(R*X + t) - b_obs on the unit sphere
+// (chord distance). Cheirality is enforced bearing-natively as
+// b_pred . b_obs > 0 (the spherical replacement for the pinhole Z(2) > 0
+// check); back-hemisphere features remain valid as long as observed and
+// predicted bearings agree on sign.
+//
+// For pinhole bearings the chord-distance residual is first-order equivalent
+// to the pixel-plane reprojection minimized by the 2D bundle_adjust above and
+// shares the same minimum in the noise-free limit, but the two objectives
+// differ by O(error^3) and produce slightly different LM iterates in
+// practice. The bearing path is the geometrically correct formulation for
+// non-pinhole central cameras.
 BundleStats bundle_adjust_bearing(const std::vector<Point3D> &bearings, const std::vector<Point3D> &X, CameraPose *pose,
                                   const BundleOptions &opt = BundleOptions(),
                                   const std::vector<double> &weights = std::vector<double>());
@@ -96,10 +104,16 @@ BundleStats refine_relpose(const std::vector<Point2D> &x1, const std::vector<Poi
                            const std::vector<double> &weights = std::vector<double>());
 
 // Relative pose refinement for any central camera model using 3D unit bearing vectors.
-// Minimizes the (x,y)-subspace Sampson error — for pinhole bearings b=(x,y,1)/|(x,y,1)|
-// this reduces algebraically to refine_relpose(Point2D, Point2D, ...) above; for spherical
-// bearings with b.z != 1 it generalizes naturally. No cheirality check — works for the
-// full sphere.
+// Minimizes the unit-norm symmetric Sampson error
+//   r = (b2^T E b1) / sqrt(|E b1|^2 + |E^T b2|^2)
+// which is the asymptotic perpendicular angular distance to the epipolar
+// great circles on the unit sphere and reduces to the standard 2D Sampson
+// for pinhole bearings (b.z = 1). The bearing residual is not algebraically
+// identical to refine_relpose(Point2D, Point2D, ...) — the pixel path uses
+// non-unit homogeneous (x, y, 1) — but the two objectives share the same
+// minimum in the noise-free limit. Cheirality is handled at the RANSAC
+// scoring step (compute_sampson_msac_score_bearing); the LM objective itself
+// does not branch on cheirality.
 BundleStats refine_relpose_bearing(const std::vector<Point3D> &bearings_1, const std::vector<Point3D> &bearings_2,
                                    CameraPose *pose, const BundleOptions &opt = BundleOptions(),
                                    const std::vector<double> &weights = std::vector<double>());
--- a/PoseLib/robust/bundle.h
+++ b/PoseLib/robust/bundle.h
@@ -48,6 +48,16 @@ BundleStats bundle_adjust(const std::vector<Point2D> &x, const std::vector<Point
                           const BundleOptions &opt = BundleOptions(),
                           const std::vector<double> &weights = std::vector<double>());
 
+// Absolute pose refinement for any central camera model (pinhole, spherical,
+// fisheye, ...) using 3D unit bearing vectors instead of 2D normalized pixels.
+// Minimizes the 3D residual r = normalize(R*X + t) - b_obs on the unit sphere.
+// No cheirality check — works for the full sphere (back-hemisphere points are
+// valid). For pinhole bearings this is algebraically equivalent to the 2D
+// bundle_adjust above.
+BundleStats bundle_adjust_bearing(const std::vector<Point3D> &bearings, const std::vector<Point3D> &X, CameraPose *pose,
+                                  const BundleOptions &opt = BundleOptions(),
+                                  const std::vector<double> &weights = std::vector<double>());
+
 // Point+Line refinement, assumes calibrated pinhole camera
 BundleStats bundle_adjust(const std::vector<Point2D> &points2D, const std::vector<Point3D> &points3D,
                           const std::vector<Line2D> &lines2D, const std::vector<Line3D> &lines3D, CameraPose *pose,
@@ -84,6 +94,15 @@ generalized_bundle_adjust(const std::vector<std::vector<Point2D>> &x, const std:
 BundleStats refine_relpose(const std::vector<Point2D> &x1, const std::vector<Point2D> &x2, CameraPose *pose,
                            const BundleOptions &opt = BundleOptions(),
                            const std::vector<double> &weights = std::vector<double>());
+
+// Relative pose refinement for any central camera model using 3D unit bearing vectors.
+// Minimizes the (x,y)-subspace Sampson error — for pinhole bearings b=(x,y,1)/|(x,y,1)|
+// this reduces algebraically to refine_relpose(Point2D, Point2D, ...) above; for spherical
+// bearings with b.z != 1 it generalizes naturally. No cheirality check — works for the
+// full sphere.
+BundleStats refine_relpose_bearing(const std::vector<Point3D> &bearings_1, const std::vector<Point3D> &bearings_2,
+                                   CameraPose *pose, const BundleOptions &opt = BundleOptions(),
+                                   const std::vector<double> &weights = std::vector<double>());
 
 // Relative pose refinement. Minimizes Tangent Sampson error error.
 BundleStats refine_relpose(const std::vector<Point2D> &x1, const std::vector<Point2D> &x2, ImagePair *pair,
--- a/PoseLib/robust/optim/absolute.h
+++ b/PoseLib/robust/optim/absolute.h
@@ -243,6 +243,101 @@ class PinholeAbsolutePoseRefiner : public RefinerBase<CameraPose, Accumulator> {
     typedef CameraPose param_t;
 };
 
+// Bearing-vector variant of the pinhole absolute pose refiner.
+// Inputs are observed unit bearing vectors (any central camera model) paired with
+// 3D world points. The residual is the 3D difference between observed bearing and
+// the normalized direction of the transformed 3D point:
+//   Z     = R*X + t
+//   b_pred = Z / |Z|
+//   r      = b_pred - b_obs        (3D residual)
+//
+// Jacobian derivation:
+//   d(Z/|Z|)/dZ = (I - b_pred * b_pred^T) / |Z|
+// which is the tangent-space projector at b_pred scaled by 1/|Z|. It is rank 2
+// (the normal-direction DOF is constrained by |b|=1), which LM handles naturally
+// through the normal equations.
+//
+// Unlike the pinhole refiner there is no Z(2)<0 cheirality check — spherical
+// cameras observe the full sphere and back-hemisphere points are valid.
+template <typename ResidualWeightVector = UniformWeightVector, typename Accumulator = NormalAccumulator>
+class BearingAbsolutePoseRefiner : public RefinerBase<CameraPose, Accumulator> {
+  public:
+    BearingAbsolutePoseRefiner(const std::vector<Point3D> &bearings, const std::vector<Point3D> &points3D,
+                               const ResidualWeightVector &w = ResidualWeightVector())
+        : b(bearings), X(points3D), weights(w) {
+        this->num_params = 6;
+    }
+
+    double compute_residual(Accumulator &acc, const CameraPose &pose) {
+        const Eigen::Matrix3d R = pose.R();
+        for (int i = 0; i < b.size(); ++i) {
+            const Eigen::Vector3d Z = R * X[i] + pose.t;
+            const double norm_Z = Z.norm();
+            if (norm_Z < 1e-12)
+                continue;
+            const Eigen::Vector3d b_pred = Z / norm_Z;
+            const Eigen::Vector3d res = b_pred - b[i];
+            acc.add_residual(res, weights[i]);
+        }
+        return acc.get_residual();
+    }
+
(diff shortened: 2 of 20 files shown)


PoseLib/robust/optim/absolute.h
```diff
@@ -243,6 +243,100 @@ class PinholeAbsolutePoseRefiner : public RefinerBase<CameraPose, Accumulator> {
...
+//   b_pred = Z / |Z|
+//   r      = b_pred - b_obs        (3D residual)
+//
+// Jacobian derivation:
+//   d(Z/|Z|)/dZ = (I - b_pred * b_pred^T) / |Z|
+// which is the tangent-space projector at b_pred scaled by 1/|Z|. It is rank 2
+// (the normal-direction DOF is constrained by |b|=1), which LM handles naturally
+// through the normal equations.
+//
+// Unlike the pinhole refiner there is no Z(2)<0 cheirality check — spherical
```
vlarsson: You can still measure cheirality for spherical cameras. In this case you just need to check whether dot product between `b_pred` and `b_obs` is positive

PoseLib/robust/optim/relative.h
```diff
@@ -165,6 +165,130 @@ class PinholeRelativePoseRefiner : public RefinerBase<CameraPose, Accumulator> {
...
+        : b1(bearings_1), b2(bearings_2), weights(w) {
+        this->num_params = 5;
+    }
+
+    double compute_residual(Accumulator &acc, const CameraPose &pose) {
+        Eigen::Matrix3d E;
+        essential_from_motion(pose, &E);
+
+        for (size_t k = 0; k < b1.size(); ++k) {
+            const double C = b2[k].dot(E * b1[k]);
```
vlarsson: I am not sure I understand this error. Is there a geometric interpretation if `b` is not normalized such that `b[2] = 1`?

I wonder if it would not be better to go directly for symmetric cosine error, something like
````
r1 = b2.T * E * b1 / norm(E * b1)
r2 = b2.T * E * b1 / norm(E.T * b2)
```

PoseLib/robust/bundle.h
```diff
@@ -48,6 +48,16 @@ BundleStats bundle_adjust(const std::vector<Point2D> &x, const std::vector<Point
                           const BundleOptions &opt = BundleOptions(),
                           const std::vector<double> &weights = std::vector<double>());
 
+// Absolute pose refinement for any central camera model (pinhole, spherical,
+// fisheye, ...) using 3D unit bearing vectors instead of 2D normalized pixels.
+// Minimizes the 3D residual r = normalize(R*X + t) - b_obs on the unit sphere.
+// No cheirality check — works for the full sphere (back-hemisphere points are
+// valid). For pinhole bearings this is algebraically equivalent to the 2D
+// bundle_adjust above.
```
vlarsson: From what I understand this optimization is not equivalent to regular pinhole BA when `b` is normalized such that third coordinate is one? Afaiu you are optimizing

```
Z = R*X+t
z = Z / |Z|
r = b_obs - z
```
while pinhole would be
```
Z = R*X+t
z = Z / Z[2]
r = b_obs - z
```
?

Or am I missing something here?

PoseLib/types.h
```diff
@@ -124,6 +125,14 @@ struct AbsolutePoseOptions {
     // and not on the actual image size.
     // Setting to 0 (or negative) disables checking.
     double min_fov = 5.0; // circa 500mm lens 35mm-equivalent
+
+    // Set max_error from an angular threshold in radians, as used by the
+    // bearing-vector pose estimator (estimate_absolute_pose_bearings).
+    // The bearing estimator scores with the squared chord distance
+    //   |b_obs - b_pred|^2 = 2 - 2*cos(angle)
+    // so the inlier threshold in those units is chord = 2*sin(angle/2).
+    // For small angles chord ≈ angle (the two agree to ~1e-4 at 5°).
```
vlarsson: Maybe the interface would be better if we just always specify max_error in terms of angle when used in the bearing estimators. And then these could internally convert to chordal etc. Would not need this helper function here as well.

vlarsson (commented):
I think it could be useful to have a pure bearing/geometry estimators as well. There are some questionmarks left though. 

1. The choice of residual functions in optimization, mainly for the two-view case. This needs to be given some thought.
2. You can still define meaningful cheirality checks for bearing-only estimators (checking positive scalar product).

I left some other comments throughout. It would be good to also add some benchmarking, e.g. via posebench to show that it works.

PoseLib/robust/optim/absolute.h
cdcseacave: Fixed in 7fb2ac2. Added bearing-native cheirality (`b_pred . b_obs > 0`) in `BearingAbsolutePoseRefiner::compute_residual` / `::compute_jacobian`, in `compute_msac_score_bearing` (skipped in the loop; the final outlier term still penalizes once), and in `get_inliers_abs_bearing`. Also updated the surrounding comments — they had been overstating "no cheirality on the sphere" when really only the `Z(2) < 0` form is wrong; the dot-product form is the proper bearing-native replacement. All 87 unit tests pass.

PoseLib/robust/optim/relative.h
cdcseacave: Fixed in d517c38. Switched the residual to full-norm symmetric Sampson:
`r = (b2·E·b1) / sqrt(|E·b1|² + |E^T·b2|²)`.

Considered both this single-Sampson form and your literal proposal of two separate residuals `r1 = C/|E·b1|, r2 = C/|E^T·b2|` and chose single-Sampson. Tradeoffs we weighed:

**Single-Sampson `C / sqrt(|Eb1|² + |E^T b2|²)` (chosen)**
*Pros:*
- First-order optimal in the linearized geometry — Sampson is the Pythagorean combination of the two perpendicular angular distances, which matches the maximum-likelihood weighting of the symmetric reprojection error to first order.
- One residual per pair → smaller normal-equation system, single scalar-vs-`Vector5d` Jacobian, ~half the per-iteration LM cost vs the two-residual form.
- Drop-in shape match with the existing `PinholeRelativePoseRefiner` (single Sampson scalar) — minimizes diff and matches the convention established for the pixel path.
- Reduces *exactly* to the standard 2D Sampson once `b.z = 1` (after making bearings unit), which keeps pinhole-bearing parity (the new test `test_pinhole_bearing_parity_relative` relies on this).

*Cons:*
- Sampson approximation only matches the geodesic angular error to O(angle³); for very large reprojection errors the cost surface differs from the true sum-of-angular-distances.
- The residual magnitude has units of `sin(angle)`, so for callers thinking in radians the `max_error = sin(angle)` conversion (now done internally inside `estimate_relative_pose_bearings`) is the bridge — slightly less direct than two-residual.

**Two-residual `C/|Eb1|, C/|E^T b2|` (your proposal)**
*Pros:*
- Each residual is exactly `sin(angle to epipolar great circle)` — direct geometric meaning, no Pythagorean approximation.
- Threshold in radians maps 1:1 to each residual via `sin(angle)`.
- Robust loss / weighting can be applied per-residual (per-image), which is closer to the symmetric-cosine literature.

*Cons:*
- Doubles the residual count; LM does ~2× work per iteration.
- For pinhole bearings does *not* reduce to the existing 2D Sampson — would diverge from the pixel-path code shape and break the parity test unless we special-case pinhole.
- Two residuals built from the same numerator `C` are correlated; LM treats them as independent which is mildly suboptimal.

Both are sound generalizations to spheres. Picked single-Sampson for the optimization-cost / pinhole-parity / code-shape symmetry; happy to switch to the two-residual form if you'd prefer the cleaner geometric interpretation.

Implementation details:
- New residual + redo analytical Jacobian in `BearingRelativePoseRefiner::compute_residual` / `::compute_jacobian` (`PoseLib/robust/optim/relative.h`). With `u = E*b1, v = E^T*b2`:
  `dr/dE(i,j) = b1(j)*b2(i)/sqrt(D) - (C/D^(3/2)) * (u(i)*b1(j) + v(j)*b2(i))`
  All 9 components of `vec(E)` now contribute (no more skipped corrections).
- Matching update in `compute_sampson_msac_score_bearing` and `get_inliers_rel_bearing` so RANSAC scoring and LM minimize the same objective.
- The existing `test_bearing_relative_pose_jacobian` (analytical-vs-FD) still passes, validating the new Jacobian.

PoseLib/robust/bundle.h
cdcseacave: You are right — they are not algebraically equivalent. Pinhole BA minimizes `(Z/Z[2]) - x_obs` (image-plane reprojection); `bundle_adjust_bearing` minimizes `(Z/|Z|) - b_obs` (chord distance on the unit sphere). These differ by O(error³) and share the same minimum only in the noise-free limit. Fixed the misleading "algebraically equivalent" wording in e6f0440 across `bundle.h`, `optim/absolute.h`, `robust.h`, and the `BearingAbsolutePoseEstimator` / `BearingRelativePoseEstimator` docstrings — they now describe the bearing residual as the chord distance / unit-norm Sampson and explicitly call out the first-order equivalence (not algebraic). The bearing path is kept as-is — using `Z / Z[2]` would defeat the purpose for non-pinhole central cameras.

PoseLib/types.h
cdcseacave: Done in 56dfef3. `opt.max_error` is now interpreted as an angular threshold in radians by both bearing entry points:
- `estimate_absolute_pose_bearings` converts internally to chord = `2 * sin(angle / 2)` before passing to `compute_msac_score_bearing` and the inlier selection.
- `estimate_relative_pose_bearings` converts internally to `sin(angle)` (the residual unit of the new full-norm symmetric Sampson — see d517c38).

Both `SetMaxErrorFromAngle` helpers on `AbsolutePoseOptions` / `RelativePoseOptions` are deleted. The `max_error` field carries a single inline comment documenting the bearing-vs-pixel interpretation. Existing tests migrated; `1e-4 rad ≈ 1e-4 chord ≈ sin(1e-4)` so the test thresholds remain valid in the new units.

cdcseacave:
@vlarsson — addressed all four unresolved comments + the benchmarking ask. Five focused commits on top of the existing PR:

| Commit | Item | Files |
|---|---|---|
| e6f0440 | Docs: drop overclaim of BA / pixel-Sampson algebraic equivalence (chord vs Z[2] reprojection) | `bundle.h`, `optim/absolute.h`, `robust.h`, `estimators/{absolute,relative}_pose.h` |
| 7fb2ac2 | Bearing-native cheirality `b_pred · b_obs > 0` in refiner + MSAC scorer + inlier selection | `optim/absolute.h`, `utils.{h,cc}` |
| d517c38 | Relative-pose Sampson: switch from (x,y)-subspace to full-norm symmetric form `r = C / sqrt(\|Eb1\|² + \|E^T b2\|²)`; redo analytical Jacobian | `optim/relative.h`, `utils.cc` |
| 56dfef3 | API: `opt.max_error` is an angular threshold in radians for the bearing entry points; conversion to chord / `sin(angle)` happens inside `estimate_*_bearings`. `SetMaxErrorFromAngle` helpers deleted | `types.h`, `robust.cc`, `tests/optim_bearing_test.cc` |
| f6cf514 | Pinhole↔bearing parity tests: synthesize a pinhole scene, drive both `estimate_*` and `estimate_*_bearings`, assert pose recovery to 1e-5 rad / 1e-5 t (abs), 1e-4 rad / 1−1e-6 cos-sim (rel) | `tests/optim_bearing_test.cc` |

All 87 unit tests pass on macOS Clang 17 (was 85; +2 parity tests). Per-thread replies on each of your comments above explain the design choices (notably the single-Sampson vs two-residual tradeoff for V2).

Posebench-proper integration deferred to a follow-up PR in `PoseLib/posebench` (it's a separate repo and would need pyposelib bindings for the bearing estimators); the in-tree parity tests cover the core "show that it works" check meanwhile.

Happy to iterate further — if you'd prefer the two-residual symmetric epipolar form for V2 over the chosen single-Sampson, or want a different threshold convention, say the word and I'll switch.

cdcseacave:
Rebased onto master at 1758bd6 (#203, #205, #207, #208, #209, #210, #211). One conflict, at the tail of `robust.cc` where #209 appended `estimate_absolute_pose_pnplf` next to `estimate_relative_pose_bearings`; both kept. The rebased diff against master is line for line the previous diff against the old base. 93 / 93 tests pass.

PR #208, opened 2026-06-12
Implementation of unknown one-sided focal length solver

vlarsson:
Implements the relative pose solver with unknown focal length for one camera, i.e. F = E * K.

Reference:
Kukelova et al., A Clever Elimination Strategy for Efficient Minimal Solvers, CVPR 2017


B1ueber2y (approved):
Thank you!

PR #210, opened 2026-07-18
Make -Werror -Wfatal-errors optional via WERROR option

### LINKED ISSUE #177: Add option to disable `-Werror -Wfatal-errors`
When updating this port @ vcpkg (https://github.com/microsoft/vcpkg/pull/49444) the arm64 CI failed with:
```
/Users/vcpkg/Data/b/poselib/src/v2.0.5-a2ce0bbe01.clean/PoseLib/misc/colmap_models.cc:644:21: fatal error: unused variable 'x0' [-Wunused-variable]
  644 |     Eigen::Vector2d x0;
      |                     ^~
``` 

Reason: #146 introduced `-Werror -Wfatal-errors` (before it was only active in case `MARCH_NATIVE` was `ON`). This flag is nice for the own CI (so you know your compiler), but should be not the default for unknown compiler or unknown compiler versions. Therefore there should be an option to disable it.

azuriru3:
The `-Werror -Wfatal-errors` flags added in #146 are hardcoded, so anyone building on a compiler version the CI doesn't test against gets a hard failure instead of a warning. That's what's blocking the vcpkg port bump and the macOS builds mentioned in #177.

Added a `WERROR` option, default ON, so nothing changes for the existing CI or `MARCH_NATIVE` builds. Pass `-DWERROR=OFF` to build with warnings as warnings instead of warnings as errors.

Checked the flag toggle produces the right compile options in both states with a small standalone CMake project (didn't have the full Eigen/dependency chain set up locally to do a real build):

```
WERROR=ON:  -O3 -Wall -fPIC -Wno-sign-compare -Werror -Wfatal-errors
WERROR=OFF: -O3 -Wall -fPIC -Wno-sign-compare
```

Fixes #177


vlarsson (approved):
lgtm

